PR #136, opened 2011-08-24
Optgroups within options

mikejolley:
When using a few optgroups within other options the optgroup headings did not behave correctly. It would select all options beyond the end of the optgroup.

I encountered this when doing a state/country selector - not all countries have states.

To get around this issue I check for the last 'option' in the optgroup and add a list element to mark the end.


mlh758:
I used a similar method to this to fix the issue, but I decided to add a class to all the children of the option group to simplify the selector somewhat.

PR #152, opened 2011-09-22
noneSelectedText, button a tag

exu:
- Allow to define noneSelectedText  as callback function.
- Change button tag to 'a' element, this allow to use tool tips and other stuff inside the button element
  (inside button element events are blocked)


mlh758:
I'm going to close this pull due to the age and merge conflicts from that.

PR #209, opened 2012-01-22
added support for lazy load of multiselect data from ajax/callback

dobozysaurus:
This pull request adds support for loading multiselect data lazily, either via AJAX or a callback function. The data is loaded once the multiselect is clicked on for the first time. The data format is assumed to be JSON, although it would be relatively straightforward to add support for XML datasources as well.

I've added qunit tests for the new functionality as well. Let me know if there's anything that's missing. I've also added a lib - mockjax - which I use to mock ajax calls on the unit tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
--- a/jquery.multiselect.css
+++ b/jquery.multiselect.css
@@ -21,3 +21,6 @@
 
 /* remove label borders in IE6 because IE6 does not support transparency */
 * html .ui-multiselect-checkboxes label { border:none }
+
+/* loading spinner */
+.ui-multiselect-loading { background: white url('images/ui-anim_basic_16x16.gif') right center no-repeat; }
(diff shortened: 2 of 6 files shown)


mlh758:
@dobozysaurus I've been going through the pull requests looking to merge in useful changes. Is this something you would still like to bring into the library? I can definitely see a use case for this.

SteveTheTechie:
I think this needs more discussion.   It seems like the OP is wanting to use AJAX implementation code that may duplicate available jQuery functionality.  

Additionally, if the multi-select is being created from JSON, does that imply that an underlying native select is not needed?   If a select is still needed, then the native select would need to be updated from the JSON, as well.

mlh758:
Yeah I've been sitting on this one for a long time because I like the idea of being able to use Ajax or some other callback to provide data after page load.

`select2` has a feature like this where you provide an Ajax call to make, and it expects the reply to be a json of values and displays. In our case, we could use that to build options into the underlying select and then call `refresh`.  I would prefer not to abandon the select.

mlh758:
I'm going to go ahead and close this one since we're discussing implementing it in version3.

PR #239, opened 2012-03-19
Custom menu size

afortaleza:
I ran into a problem where I need a small select box but a bigger menu size, there was not an option to set the menu size so I have added one, when this option is set the menu respects the size set in the option, otherwise it behaves the original way, following the select box size.


mlh758:
It looks like menuWidth was added already at some point since this was opened.

PR #240, opened 2012-03-25
fix blank option display

alexeyramazanov:
Sometimes it is needed to add blank option in select list. Currently it shows not very good.
before
This patch fixes display issues.
after


mlh758:
Closing this PR due to age, limited use case, and merge conflicts.

If you still want to see this in the library you can open a new PR and we can figure it out from there.

PR #242, opened 2012-04-13
Allow to set specific menuWidth by option

doits:
For specific designs it is nice to set a specific menu width.


mlh758:
This is essentially a duplicate of the recently merged PR #567

PR #245, opened 2012-04-14
Group different <optgroup>s in different lists

doits:
This allows more custom layouts, like multiColumn-lists.

screencap

There's some additional css needed for multiColumn-lists, but I could provide it if you want it as a feature of this plugin.

This pull-request just lays the ground work and changes things internally, without (hopefully) changing the behavior.

If you consider this valid and mergable, I'll write some test cases.

### DIFF
--- a/src/jquery.multiselect.filter.js
+++ b/src/jquery.multiselect.filter.js
@@ -27,9 +27,10 @@
 
     _create: function() {
       var opts = this.options;
+      var elem = $(this.element);
 
       // get the multiselect instance
-      var instance = (this.instance = $(this.element).data('multiselect'));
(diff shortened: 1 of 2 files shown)


jordangurney:
Thank you so much for this - this is exactly what I need for an access control list - now I just need to figure out how to implement it...

mlh758:
I recently rewrote the `refresh` method, which makes the merge for this especially difficult. It looks like the gist of this request is to group all the children of the `optiongroup` tags under a `ul` so they can get their own styling?

That may be a lot easier to implement now with the new `refresh` method.

mlh758:
This has been incorporated and pushed into master. The `groupColumns` option will display option groups as columns instead of rows.

PR #254, opened 2012-05-07
another IE 7/8 JS error fix (Cannot move focus to control because it is invisible/disabled)

lion19:
Note: this request helps with a bug encountered in ie8 with jQuery 1.7.2


mlh758:
I'm closing this due to how stale this has gotten.

PR #308, opened 2012-09-20
Update src/jquery.multiselect.js

ericpanorel:
when using knockout with an array objects bound to the options
instead of an array of strings, ko doesn't set the value attribute.
This patch compensates that by doing a coalesce 
(this.value ? this.value : this.innerHTML)


mlh758:
Closing this PR since it has become so out of date with master.

PR #315, opened 2012-10-09
Added collapsible optgroups and custom checkbox (full/partial/none) optgroups

aerichmond:
Feel free to merge any or some of the changes I have made.  For large (lots of data) dropdowns I found the collapsing and checkbox of optgroups much more user friendly.

andy


src/jquery.multiselect.js
```diff
@@ -1,699 +1,887 @@
...
+					this.refresh();
+					break;
+				case 'position':
+					this.position();
+					break;
+				case 'optGroupCollapsible':
+					checkboxContainer.toggleClass('ui-multiselect-optgroup-collapsible', value);
+					break;
+				case 'optGroupSelectable':
+					checkboxContainer.toggleClass('ui-multiselect-optgroup-selectable', value);
```
ballavamsi: I think checkboxContainer should be either this.checkboxContainer  or you have to declare var in the method 
var cbContainer = this.checkboxContainer; 
and in case statement
cbContainer.toggleClass('ui-multiselect-optgroup-collapsible', value);

Correct me if I am wrong.

mlh758:
This pull request has gone pretty stale in general since it isn't compatible with the rewritten refresh method.

PR #318, opened 2012-10-16
jquery.multiselect.js: added bind to opt group that only allows user to ...

DaleKocian:
...select options associated with opt group


mlh758:
I'm closing this due to age, merge conflicts, and extremely specific use case.

PR #330, opened 2012-11-07
Add a buttonvaluechanged event 

cedriclombardot:
Easing the way to add a photo or something else in the button value after the update


mlh758:
You could also use the close event or use a `setTimeout` to make the change after the other event handlers have fired.

PR #338, opened 2012-11-20
RTL support css files and Persian(Farsi) i18n

ghasem:
To support Right-to-Left, I'm just including the rtl css file after main
css files to overwrite the Left-to-Right css(default), but it'd be
better if we have an rtl option in the widget. it's just working fine ;)


mlh758:
Closing since there haven't been any changes since the suggestions were brought up.

PR #342, opened 2012-11-22
Allow to disable titles of items

doits:
Swap variable name in code, too, to match better what they are for


mlh758:
Closing PR since this conflicts with a more recent change to the library.

PR #364, opened 2013-01-05
Show/Hide animation with effect, direction and speed

ConvMe:
Using show and hide with effect, direction and speed. E.g.

show: ["slide", {direction: 'up'}, 100],
hide: ["slide", {direction: 'up'}, 100],

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -553,13 +553,14 @@
       // figure out opening effects/speeds
       if($.isArray(o.show)) {
         effect = o.show[0];
-        speed = o.show[1] || self.speed;
+        direct = o.show[1];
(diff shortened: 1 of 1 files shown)


SteveTheTechie:
I think we should be able to pull this into version 3.  It is pretty straight forward.

SteveTheTechie:
@mlh758  I have not been able to locate where the OP's mentioned syntax for including the direction is mentioned--current syntax looks different.  (and this is a very old PR)   I did review both the plain jQuery show/hide and the jQuery UI show/hide docs.  Current syntax looks like an object is supplied.  Does this agree w/ your understanding?

mlh758:
Yeah, the API for show/hide seems to have changed a bit: http://api.jqueryui.com/show/

PR #381, opened 2013-02-02
Pop the menu over the button if it will be cut off by the bottom of the window 

TerryMooreII:
multiselect

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -671,7 +671,7 @@
         // otherwise fallback to custom positioning
       } else {
         var pos = this.button.offset();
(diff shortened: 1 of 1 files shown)


mlh758:
Does this still have a use case now that the [position option](https://github.com/ehynds/jquery-ui-multiselect-widget/wiki/Options) is available in the library?

TerryMooreII:
I have no idea.  This is 2 years old and I no longer work at that company.

mlh758:
@TerryMooreII Fair enough, I'm going to go ahead and close this then since the position option seems to handle that case.

I appreciate your submission though, and I apologize for the gap in support.

PR #428, opened 2013-04-17
Allow space to select

r4j4h:
We added the ability to select items via space, and disabled keyup events to prevent bouncing because firefox seemed to read on its own unlike the other browsers, and we are catching "real" expected events in keydown.

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -350,6 +350,7 @@
             self._traverse(e.which, this);
           break;
           case 13: // enter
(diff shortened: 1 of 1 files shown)


r4j4h:
*bump* Added the `selectOnSpace` option to control feature.

mlh758:
I've added this feature in. I wasn't able to merge the pull due to merge conflicts but I agree that space bar makes sense for selecting and it wasn't in use for anything else.

PR #436, opened 2013-04-23
Update jquery.multiselect.js

andrey-boomer:
Set dropdown menu as resizable, if Jquery UI and method resizable is presed. And save height into options.


SteveTheTechie:
I think this can be pulled into version 3.  However, the PR code as presented here is not sufficient.  The dimensions of the various components of the widget will need to be recalculated on a resize.

PR #437, opened 2013-04-23
Update jquery.multiselect.js

andrey-boomer:
Is needed to rotate menu div to a selected line, but worked only in single select mode.


mlh758:
I believe our updated use of the position utility should cover this.

PR #441, opened 2013-04-24
Update jquery.multiselect.js

ecappa:
EC - selectedList small issue: a text() is better than hmtl... in my case, html artefacts were generated (&nbsp; for eg.)


mlh758:
I merged #442 instead of this pull to avoid extra comments in the code. They were otherwise identical.

Issue #443, opened 2013-04-25
Pass separator string as option?


abennouna:
Great plugin that I have just been able to use :-)

I happen to use it in a bi-directional app, and the comma currently used to separate the displayed checked (selected) values should adapt to the language.

The line 225 seems to be the one:
`value = $checked.map(function() { return $(this).next().html(); }).get().join(', ');`

To be as generic as possible, do you think it is possible to specify the separator as an option?

In the case I'm not clear: here's what the comma looks in Arabic:
`،`

So multiple selected values can be:
`قط، فأر`

which is Arabic for

`cat, dog`

mlh758:
There is an option available for this: https://github.com/ehynds/jquery-ui-multiselect-widget/wiki/Options

`selectedListSeparator`

PR #444, opened 2013-04-25
Adding an option for the items separator string

abennouna:
This is especially useful in bidi apps, but also when you want to
specify a separator string different from the default "comma-space"

(related to issue https://github.com/ehynds/jquery-ui-multiselect-widget/issues/443)


mlh758:
Closing since this is an existing option: `selectedListSeparator`

If you want to open a new pull request for the i18n files that would be helpful.

PR #458, opened 2013-05-22
getGroup: new method

supermarcos:
Hi, I have been added a new method getGroup to get the group of the checked option by the input check box.
If you see it can be improve, just do it, please. Everything always can be improved.
Thank you!


mlh758:
This would need to get rewritten somewhat to work with the redesigned refresh method. I'm going to close it for now.

PR #465, opened 2013-05-31
Group Features - Collapsible and Filterable

easytomanage:
This implements two improvements for groups.

1) For the standard multiselect, I added a collapse / expand icon near each group. This causes the group to collapse, yet clicking on it in it's collapsed state will check / uncheck its children.
2) For the filter plugin, I added support to search by group name. If you type a group name, or part of it, all of its children will be displayed, even if the checkbox itself does not match.

### DIFF
--- a/demos/groups.htm
+++ b/demos/groups.htm
@@ -1,0 +1,68 @@
+﻿<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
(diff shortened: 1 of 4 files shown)


mlh758:
I'm closing this since this change is no longer compatible with the 2.0 widget. However, I think the new mechanism for displaying the list should simplify the implementation of collapsing groups. If you still want this integrated, I would happily merge your changes if they were brought up to date.

PR #476, opened 2013-07-09
Empty text in ui object

maintux:
When you use callbacks and option tags doesn't have a title attribute, the field called 'text' in the ui object is empty. With this commit, if title attribute is empty, the text field is filled with the text of option tags.


mlh758:
This was resolved in PR #636 but I appreciate you sending this along. Had I known it was here I would have merged it before making my own changes.

PR #442, opened 2013-04-25
Prevent double-encoding of HTML entities in the widget's button text value

Data-Dev-Team:

PR #480, opened 2013-08-06
Prevent html encoding artifacts

quarksoup:
Change .html() to .text() so that special html characters (ie, '<', '>') are not encoded when using this control in single select mode.


mlh758:
#442 Provided the same code change, I have merged that pull so I'm going to close this one.

PR #538, opened 2014-02-28
make close icon configurable

kielni:
The default close icon is very close to the clear icon, which can be confusing.  I've added an option to set the close icon to something else (I'm using ui-icon-search).


mlh758:
This looks good. I appreciate the addition of the unit test.

PR #541, opened 2014-03-07
Selecting from multiple dropdowns causing problems

### LINKED ISSUE #508: Selecting from multiple dropdowns causing problems
Hi,

I'm creating multiple drop downs on one page, and when I select a checkbox on any of the dropdowns, i get the expected behavior, but if i click on the name of the country instead of the box then the first dropdown on the page appears to get that country selected instead of the one i am trying to edit.

Do you know what could be causing this?

Thank you!

pawel-piskorz:
- updated inputID to include multiselectID in order to be unique
  across all select with the same name (Fix #508)

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -139,7 +139,7 @@
         var description = this.innerHTML;
         var title = this.title;
         var value = this.value;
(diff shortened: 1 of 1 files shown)


mlh758:
Looks good, I'll merge this in.

PR #542, opened 2014-03-07
Allow no set default width with minWidth: "none"

davehermann:
The current behavior allows for setting the width of both the button and the menu div using a stylesheet and jQuery.outerWidth().  Not specifying a width (i.e. minWidth is "auto") yields 0px for both elements.

This allows for an unspecified width, letting browsers wrap horizontally, using
    minWidth: "none"

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -427,18 +427,22 @@
       var width = this.element.outerWidth();
       var o = this.options;
 
-      if(/\d/.test(o.minWidth) && width < o.minWidth) {
-        width = o.minWidth;
-      }
+      if (o.minWidth != "none") {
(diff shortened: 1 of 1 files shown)


mlh758:
Closing since so much of the behavior around width has changed.

davehermann:
After 4 years, that seems reasonable to me.  I moved on from jQuery UI quite some time ago, and no longer work for the company where the widget was in-use.

Thanks.

mlh758:
Well the widget wasn't being supported for a long time, then work was done to fix a bunch of the issues in the backlog and make other updates, and I've finally started taking a look at the PRs.

I'm leaving some open that I might try to re-implement as features but closing others that either conflict with or do similar things to changes that have already been made. I figured you would appreciate a message for why it was closed rather than suddenly seeing a closed PR 4 years later with no explanation.

PR #561, opened 2014-06-17
Update to make this work with jqGrid

kccarter76:
I have utilized this control to plug into the jq grid as a custom control and the changes made where to make this widget look like it belongs and interops with the jq grid.

### DIFF
--- a/jquery.multiselect.css
+++ b/jquery.multiselect.css
@@ -9,11 +9,11 @@
 .ui-multiselect-header a { text-decoration:none }
 .ui-multiselect-header a:hover { text-decoration:underline }
 .ui-multiselect-header span.ui-icon { float:left }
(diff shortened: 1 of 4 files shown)


mlh758:
It's difficult to see what the actual changes are for this due to all the white space changes. If you're still interested in this, please fix the white space issues [and squash this into one commit.](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit)

kccarter76:
I no longer work for the company, in which the filter extension was being used so I no longer have access to that source file and it looks like github has made a mess of it. I am going to close since I can't remember what the actual changes were.

mlh758:
Fair enough, sorry about that. I'm going to close your other open pull then as well since it pertains to the same thing.

PR #565, opened 2014-07-15
Adds a toggleInputs option to the enable/disable methods

Dan-Nolan:
For select elements with a large number of options, Internet Explorer will freeze when enabling and disabling the multiselect. After looking at the code it appears that all of the option elements are being enabled and disabled as well. Since the main button is disabled it doesn't appear that the option elements always need to be disabled/enabled. For some implementations this may be an unnecessarily large DOM manipulation. 

Here's an example:
http://jsfiddle.net/akhyp/1496/

In Internet Explorer this could freeze the window for upwards of 30 seconds, just trying to disable the element. 

This fork allows the developer to choose whether or not we need to the toggle the inputs inside of the select element, which for us makes all the difference between using the disable functionality or rolling our own. The default behavior of toggling these elements is preserved, but the developer can additionally pass in a "falsey" value to opt out of toggling the inputs.

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -515,25 +515,28 @@
       }
     },
 
-    _toggleDisabled: function(flag) {
+    _toggleDisabled: function(flag, toggleInputs) {
       this.button.attr({ 'disabled':flag, 'aria-disabled':flag })[ flag ? 'addClass' : 'removeClass' ]('ui-state-disabled');
 
-      var inputs = this.menu.find('input');
-      var key = "ech-multiselect-disabled";
 
-      if(flag) {
-        // remember which elements this widget disabled (not pre-disabled)
-        // elements, so that they can be restored if the widget is re-enabled.
-        inputs = inputs.filter(':enabled').data(key, true)
-      } else {
-        inputs = inputs.filter(function() {
-          return $.data(this, key) === true;
-        }).removeData(key);
+      if(toggleInputs) {
+        var inputs = this.menu.find('input');
+        var key = "ech-multiselect-disabled";
+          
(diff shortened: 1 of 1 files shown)


src/jquery.multiselect.js
```diff
@@ -616,12 +619,12 @@
...
-    enable: function() {
-      this._toggleDisabled(false);
+    enable: function(toggleInputs) {
+      this._toggleDisabled(false, toggleInputs === undefined ? true : toggleInputs);
     },
 
-    disable: function() {
-      this._toggleDisabled(true);
+    disable: function(toggleInputs) {
+      this._toggleDisabled(true, , toggleInputs === undefined ? true : toggleInputs);
```
mlh758: Should there be two commas here?

src/jquery.multiselect.js
```diff
@@ -616,12 +619,12 @@
       this._trigger('close');
     },
 
-    enable: function() {
-      this._toggleDisabled(false);
+    enable: function(toggleInputs) {
+      this._toggleDisabled(false, toggleInputs === undefined ? true : toggleInputs);
```
mlh758: It might be easier (more readable) to just check the parameter in the function instead of using a ternary operator in the function call.

mlh758:
This whole thing is kind of old, but I wanted to provide you an update.

In a [recent commit](https://github.com/ehynds/jquery-ui-multiselect-widget/commit/4ee469b87c81ba2c78ab5831d3f285dd1fe5f8d2) I added the option you requested.

I also redesigned the disable functionality and it runs substantially faster, to the point where the option probably isn't necessary. The toggle takes less than a second.

Issue #551, opened 2014-04-16
click to scroll issue


metinata:
Hello,

I have a horizontal scrolling page and when click selectbox scrollbar going to left side

SteveTheTechie:
@mlh758 This should be checked against the new code.

mlh758:
Checking the PR it looks like I wasn't able to verify the issue in IE 11 or Edge even on the 2.0 code and never closed these issues.

Given the dates this was probably an older IE problem. I'll still double check though.

SteveTheTechie:
@mlh758 This should have auto-closed, but did not.

PR #574, opened 2014-09-23
Update jquery.multiselect.js to avoid jump scroll

joshdance:
Steps to reproduce the bug that this update fixes. - https://github.com/ehynds/jquery-ui-multiselect-widget/issues/551

Load page with SingleSelect element - for example - http://www.erichynds.com/examples/jquery-ui-multiselect-widget/demos/#single
Change browser window to very narrow
Scroll over to the left hand side.
Click single select element
Observe the page jumps back to the right hand side, or right aligned.
Screencast of bug on demo page - http://screencast.com/t/GUmkvIQzIg

Screencast of bug in our system (all info is standard fake Salesforce info) - http://screencast.com/t/xL3buTWvZ

This doesn't change behavior, keyboard events still work, focus is still given, but the page does not jump around.

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -22,6 +22,14 @@
 
   var multiselectID = 0;
   var $doc = $(document);
+  
+  function focusNoScroll(multiselectObject) {
+    //save the current window position, and restore it after the focus event. 
+    var x = window.scrollX, y = window.scrollY;
+    multiselectObject.trigger("focus");
+    window.scrollTo(x, y);
+    return multiselectObject; //chainability
+};
 
   $.widget("ech.multiselect", {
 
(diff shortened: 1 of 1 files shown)


debugger01:
the update solves the issue, screen stop jump on mouse hover in the list menu.

but it jump on value selected.

Taking your screencast as eg:

Screencast of bug in our system (all info is standard fake Salesforce info) - http://screencast.com/t/xL3buTWvZ

 Imagine they are a few Select Lists on the right (same column with Preferred Contact Method), after user input one of them, it jump to left, user has to scroll back to right to fill out the second Select List and so on.

Do you have solution for this? 

Thanks!

joshdance:
@debugger01 I don't have a solution. And I no longer work on Salesforce apps. Sorry. Good luck.

mlh758:
I know this has been a while, but can you tell me what version of IE this was happening in? Following the steps in the screen casts I haven't been able to reproduce the jump using IE 11 or Edge.

PR #579, opened 2014-10-27
Allow user to hide either select all / select none

sgonzalez3884:
Allow user to set either checkAllText or uncheckAllText to null/false to suppress display in header


mlh758:
This feature exists already.

PR #584, opened 2014-11-17
Pre check alland check none events

r4j4h:
I needed my checkAll/checkNone callback to be able to handle the checkboxes as they are before checking them all or removing the checked state.


SteveTheTechie:
@mlh758 #747 addressed this.  This can now be closed.

PR #607, opened 2015-05-01
Hungarian translation

AdamFonagy:
Two translation files are added to i18n directory for Hungarian language.


AdamFonagy:
wrong header

PR #617, opened 2015-08-09
copy data-* attributes of options & optgroups

nils-a:
I needed the possibility to show/hide options dynamically and the site I'm currently working on relies somewhat on data-\* attributes instead of "functional classes".

So I needed the ability to copy data-\* attributes as well as classes from options and optgroups


mlh758:
I don't know if you're still interested in this pull request.

There's been a large refactor of the `refresh` method which makes this impossible to merge. However, a large part of this has already been added. `data-` attributes from the options are being copied to the input elements. How vital would the option group attributes be?

Issue #624, opened 2015-11-05
Up key cycles between the first and last item of the selection


FrancisMurillo:
I noticed when the menu is open and you press the up key,
the selection goes to the first item, then the last item and back again.

Good component by the way

FrancisMurillo:
I found the source of the behavior at the traverse function

```javascript
 _traverse: function(which, start) {
      var $start = $(start);
      var moveToLast = which === 38 || which === 37;
```

It seems here that the Up(38) and Left(37) button are really designed to do this behavior.
I find it a bit odd.

Is there a bug or issue that is being avoided because of this?

mlh758:
@FrancisMurillo There is a pull request open to amend this behavior. Does Left(37) and Right(39) going to top and bottom make more sense to you from a usability standpoint?

FrancisMurillo:
@mlh758 
The true concern is just the up and down button. As long as it goes up and down, then I guess it's okay.

PR #632, opened 2016-02-18
Fix keyboard navigation.

Sam152:
This amends the up/down left/right keyboard navigation.


mlh758:
Is this related to issue #624?

mlh758:
@Sam152 This looks good, my one gripe is that selecting the first item (left arrow now) no longer scrolls all the way to the top, just goes to the first point at which the element is visible. This means that if there is an option group label you won't be able to see it. If you add that in I will merge the pull request.

Sam152:
Sorry, I don't have the capacity to continue to look at this right now.

PR #636, opened 2016-03-13
Updated Unit Tests

mlh758:
Unit tests are now up to date and passing.

Index.html is referencing current libraries rather than base revisions. As per: http://stackoverflow.com/questions/12608242/latest-jquery-version-on-googles-cdn

Using the base revisions to get the latest library versions is no longer supported so we need to manually update these to the latest as they come along.

I set the title property to a sanitized version of the option elements text content if no title was initially provided. This is in line with the assertions in the unit tests.

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -137,7 +137,7 @@
         var $this = $(this);
         var parent = this.parentNode;
         var description = this.innerHTML;
-        var title = this.title;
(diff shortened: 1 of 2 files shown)


mlh758:
@gs-akhan This should fix the failing unit tests. Any thoughts?

gs-akhan:
+1

Issue #638, opened 2016-03-24
toggle select mode doesn't work


benzid-wael:
My code toggle between select modes dynamically,  when I toggle to _single_ select mode and then go back to _multiple_ select mode, I didn't see **Check all** / **Uncheck all** commands even after refreshing

PR #639, opened 2016-03-24
Fix changing select mode, #638

benzid-wael:
This update headerLinkContainer visibility when we change select mode


mlh758:
Looks good, I'll merge this in now.

Issue #229, opened 2012-02-27
xss attacks within the options


kogaion:
Hello,

I believe that, when updating the "buttonlabel" <span> with the selected option text(s), the html must first be escaped to avoid xss-attacks within the option text, for instance like this: 

$.ech.multiselect.prototype.update = function () {  
            var o = this.options, $inputs = this.labels.find("input"), $checked = $inputs.filter(":checked"), numChecked = $checked.length, value;
            if (numChecked === 0) {
                value = o.noneSelectedText;
            } else {
                if ($.isFunction(o.selectedText)) {
                    value = o.selectedText.call(this, numChecked, $inputs.length, $checked.get());
                } else if (/\d/.test(o.selectedList) &&
                    o.selectedList > 0 && numChecked <= o.selectedList) {
                    // OLD, NOT ESCAPED
                    //value = $checked.map(function () {return this.title; }).get().join(", ");
                    // NEW, ESCAPED
                    value = $checked.map(function () {return $('&lt;div /&gt;').text(this.title).html();}).get().join(", ");
                    // END OF CHANGES
                } else {
                    value = o.selectedText.replace("#", numChecked).replace("#", $inputs.length);  
                }
            }
            this.buttonlabel.html(value);
            return value;
        };

Thank you.

PR #640, opened 2016-03-29
Remove entities from the button value (Closes #424).

### LINKED ISSUE #424: with selectedlist option the ampersand "&" becomes &amp;
If I have an option in the multiselect list like "cats & dogs" 
in the selectedlist.htm demo than the character "&" (ampersand) becomes "amp;" if I select it!
?

lefterisnik:


mlh758:
This is already resolved in the latest code using a different method.

The setButtonValue method was set this way intentionally to allow developers to put HTML and potentially script tags in place if they _really_ wanted to but the default behavior was intended to be safe. https://github.com/ehynds/jquery-ui-multiselect-widget/issues/229

This feature is in the docs as well: https://github.com/ehynds/jquery-ui-multiselect-widget/wiki/Customization

mlh758:
I will update and close the issue.

PR #714, opened 2018-01-05
Eliminate unnecessary function calls

SteveTheTechie:
Eliminate unnecessary function calls to improve performance--extra addClass() calls not needed.


mlh758:
Looks good, I'll try to have this merged today.

PR #715, opened 2018-01-05
Eliminate deprecated jQuery API calls

SteveTheTechie:
Swapped out deprecated $().delegate() API calls for equivalent $().on() API calls.


mlh758:
This removes some backwards compatibility but I already did that when I started using `instance` in filter so I'm good with this. jQuery 1.7 has been around since 2011...

PR #716, opened 2018-01-05
Remove web URL for original maintainer

SteveTheTechie:
Original maintainer web URL no longer works... removed.


mlh758:
I dunno, are the unit tests still passing?

Issue #717, opened 2018-01-05
I have a number of updates...


SteveTheTechie:
Thank you for approving my pull requests.

I have been using a older version of this widget for a few years that I have made a number of updates to on my own.  The widget has had (and continues to have) a lot of potential and value.  However, because I made my changes on my version and there have obviously been changes made here, there are quite a few differences at this point.

Some of my changes I made in the older copy were intended to accomplish the following:

- [x] 1. Make the widget more performant--the original version I started with was very slow in some cases and used jQuery excessively to the point where it was killing speed.  In particular, initializing the widget successively in a large table shows just slow it is--the pop-up list generation seems to be the main culprit.   A blended "less is more" strategy that uses either more cached jQuery and/or less jQuery (more straight JavaScript) results in a faster widget.

- [x] 2. Implemented capability to automatically determine best minWidth based on underlying `<select>` options.

- [x] 3. Implemented ability to change the text used for the default links in the header or omit specific default links. This is very useful when maxSelected is in effect, since Check All will likely cause the number of selections to exceed the maximum, but Uncheck All still has value then.

- [x] 4.  Implemented "Flip All" header link that toggles the state of each check box.

- [x] 5. Changed the update routine to correctly reposition the menu (down) for a growing list of selections in the button which causes the button height to grow.

- [x] 6. Ditched the use of jQuery UI icons, in favor of similar HTML entities.  The jQuery UI icons are not scalable, whereas the HTML entities are.

- [x] 7.  Ditched the multiple option.  The multiple attribute of the underlying element is used instead in order to enforce the widget behavior to align with the select element's attributes.

- [x] 8.  Implemented capability to set the menu height based on the size attribute of the underlying select element.

- [x] 9.  Implemented a built-in maxSelected behavior and option.  If maxSelected is set to a number, the default alert message is displayed.  If it is set to a function name, the function is ran 	and behavior depends on return value.  **_Not really completed yet._**

If you are open to it, I would like to incrementally post a number of pull requests to get the two versions closer together.  I also have a number of additional enhancements I have been putting off making.

Can you tell me what your maintenance objectives are for this widget, moving forward, so that I am not spending time on updates that are not wanted?

SteveTheTechie:
The comment numbering does not seem to work as I expected here.  There are more than 3 items above.

mlh758:
If I edit your comment I can see the full list. Probably a format keyword somewhere that's messing the whole thing up. Reading through it now.

[...]

PR #719, opened 2018-01-07
Improved multiple handling

### LINKED ISSUE #712: Changing multiple option clears selected values
Conversation started in #377 breaking it out to a separate issue

At first glance this seems intentional, changing the multiple option would change the intent of the widget so clearing the options would be desired behavior. This is deliberately done in the handler for multiple.

However, @jasonperrone it seems like you're seeing a different issue. Your code is re-calling the initializer and this is clearing the values. Could you elaborate on that? Maybe provide a code snippet or a fiddle? I don't think I'm completely clear on your use case.

SteveTheTechie:
These changes fix #712 and also introduce the capability to just detect the multiple attribute from the underlying select element in lieu of explicitly setting it via an option.

In addtion, the appropriate unit tests were updated.


mlh758:
Please revert the minified file, I update those myself at release. I've just had too many issues with conflicts or encoding errors when merging minified files in from a PR.

I'll try to take a look at everything else soon. I'll also want you to squash the commits down into one before merging but I think I see some mixed tabs and spaces we'll want to fix first.

mlh758:
I pulled down the PR and started testing it a little bit. I used the single select demo, added the multiple attribute to the underlying select, and then called `refresh`. The widget didn't switch to multiple mode from single select. Are you able to reproduce that?

SteveTheTechie:
Michael, thanks for taking a look at this.  I am a bit of a git/GitHub neophyte (I do this as a hobby), so please excuse me if I do not immediately understand what you are asking me to do.  I am not sure how to revert one file out of a set.

I will take a look at your other observation.

mlh758:
Sometimes people use this widget in such a way where one item on the form will dynamically determine whether a user should pick one or many items from the drop down. In that case they would have used the `multiple` method to change the widget at run time based on user actions.

I agree that using the native select's multiple attribute makes more sense than trying to duplicate that attribute in the widget but I don't want to lose the behavior of being able to switch the widget at run time.

mlh758:
Taking a step back for a second, was the goal of this PR to remove the need for the multiple option?

Or was it to improve the checking of the native attribute when the widget is loaded/refreshed to make it more reliable?

SteveTheTechie:
Well, originally it was the first.  However, because of the likely need to maintain some sort of backward compatibility, I kept in support for the multiple option by changing its default value to null.  E.g., if you set it to an actual value of T/F, then it gets used for the sake of backwards compatibility.   

However, my recommendation would be to just deprecate use of the multiple option and just ditch it at some point.  The only reason to keep it is that you can change it and cause the widget refresh in one step--if you use the native attribute, you have to change the attribute, then call the widget refresh (2 steps).  However, I think there is a case to be made about not duplicating options that are already part of the native select element--keeping them in sync is a bit of a headache.

I added the checking of the native attribute to the refresh to support the dynamic changing of single/multiple, as you had found an issue w/ that in your prior comment and I realized that I needed to do that check to support dynamically changing single/multiple.

SteveTheTechie:
One other thing to keep in mind is that the native behavior of a single select is not the same as the widget's modeled behavior of a single select in the current code.  The native single select must always have an option selected unless you set the selectedIndex property to -1 (kludgey) which is not something that I think would typically be done.

However, the widget is really designed to work off of an underlying native select w/ multiple always set and to just model the behavior of a native single select.  I can point to evidence of this assertion. 
Because it actually an underlying multiple select, all options *can* be deselected, a distinct difference in behavior.

I ran into this difference in the uncheckall test for a single select.   I ended up implementing the selectedIndex = -1 approach in the uncheckall method to make the test work.

mlh758:
We might be able to do a best of both worlds here...

Currently it seems like the issue is that the option for multiple isn't always synced up with the underlying select. I'd rather not do this half way, so I think getting rid of the option and just basing the behavior on the underlying select would be the way to go. When we merge in all the changes you talked about in #717  that would probably justify a 3.0 release and we can deprecate some behavior.

One of us can just add a method for `multiple` that changes the underlying element and refreshes the widget similar to the existing behavior. That gives an easy way to change the widget on the fly, and a lot of the code should already exist.

Does that sound reasonable to you?

SteveTheTechie:
Ok, but I think we need to come to an agreement with how closely we want the single select version to match a native single select's behavior.    I am referring to the uncheckAll case... the native single select would not be expected to have nothing selected (goes to default value) except for the selectedIndex=-1 kludgey case.

Lastly, as it appears that we are collecting a number of changes, perhaps you may want to create a 3.0 branch so that we can keep these separate until we have all of them stabilized then you could just merge that branch to master.

mlh758:
Good call on the 3.0 branch. People shouldn't be pulling straight from master but you never know... And it'll be easier to track what we've changed.

Looking at the behavior on master, when the widget is in single select mode there isn't a button for uncheck all. Calling the method in code for `uncheckAll` manually causes the button text to change, but when I call `val()` it's still the first option. I think we shouldn't try to get too clever, and mirror the underlying select's behavior. The more popular `select2` widget appears to behave this way as well. A blank option can be handled the old fashioned way - just include a blank option as the first entry. 

It doesn't look like the current implementation is entirely consistent.

SteveTheTechie:
"It doesn't look like the current implementation is entirely consistent."

Exactly.  I also think we should also mirror the behavior of the underlying select as much as we can.

SteveTheTechie:
Even README.markdown mentions the underlying select control as being a "multiple select".

SteveTheTechie:
Closing this pull request as I intend to delete my fork and create a new fork (and pull request) based off of version3 branch.

PR #723, opened 2018-01-10
Multiple option elimination (redo of #719)

SteveTheTechie:
_(This is a re-do of #719 on the new version3 branch.)_

* Update jquery.multiselect.js

Eliminate the need for the multiple option by instead relying on the multiple attribute of the underlying native select element.

* Update core.js
* Update options.js

Update tests due to elimination of using multiple option in favor of using multiple attribute of underlying native select.

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -29,19 +29,19 @@
       height: 175,
       minWidth: 225,
       classes: '',
-      openIcon: '<span class="ui-icon ui-icon-triangle-1-s"></span>',   // Scaleable HTML Entities or Font-Awesome icons can be used here instead of the default jQuery UI icons.
+      openIcon: '<span class="ui-icon ui-icon-triangle-1-s"></span>',         // Scaleable HTML Entities or Font-Awesome icons can be used here instead of the default jQuery UI icons.
       closeIcon: '<span class="ui-icon ui-icon-circle-close"></span>',
       checkAllIcon: '<span class="ui-icon ui-icon-check"></span>',
       uncheckAllIcon: '<span class="ui-icon ui-icon-closethick"></span>',
       flipAllIcon: '<span class="ui-icon ui-icon-arrowrefresh-1-w"></span>',
-      checkAllText: 'Check all',																							// If blank or null, link not shown.
-      uncheckAllText: 'Uncheck all',																				// If blank or null, link not shown.
-      flipAllText: 'Flip all',																									// If blank or null, link not shown.
+      checkAllText: 'Check all',                                              // If blank or null, link not shown.
+      uncheckAllText: 'Uncheck all',                                          // If blank or null, link not shown.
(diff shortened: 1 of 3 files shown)


src/jquery.multiselect.js
```diff
@@ -419,15 +430,15 @@
             self._traverse(e.which, this);
             break;
           case 13: // enter
-          case 32: //space
```
mlh758: I would prefer you leave the key code comments in.

src/jquery.multiselect.js
```diff
@@ -29,18 +29,23 @@
...
       classes: '',
-      checkAllText: 'Check all',
-      uncheckAllText: 'Uncheck all',
-      noneSelectedText: 'Select options',
+      openIcon: '<span class="ui-icon ui-icon-triangle-1-s"></span>',   // Scaleable HTML Entities or Font-Awesome icons can be used here instead of the default jQuery UI icons.
+      closeIcon: '<span class="ui-icon ui-icon-circle-close"></span>',
+      checkAllIcon: '<span class="ui-icon ui-icon-check"></span>',
+      uncheckAllIcon: '<span class="ui-icon ui-icon-closethick"></span>',
+      flipAllIcon: '<span class="ui-icon ui-icon-arrowrefresh-1-w"></span>',
+      checkAllText: 'Check all',																							// If blank or null, link not shown.
```
mlh758: It looks like there's a bunch of white space here to try and line these comments up. I'm guessing tabs? It's not consistently lined up on my editor, probably a tab width thing. I'd prefer you just stick with spaces and put them next to the options.

mlh758:
I'd also like you to change your commit messages. You have `#1 #2 #3` in the messages and those are linking to issues 1,2, and 3 in the issue tracker of this repo which I don't think is intended.

src/jquery.multiselect.js
SteveTheTechie: I think this was likely an unintended accidental change on my part.   I generally prefer more comments, so removing them is likely a goof.   Sorry, I will fix.

src/jquery.multiselect.js
SteveTheTechie: Yeah, I am fighting with the tab vs space thing.   Generally, I prefer tabs, but I realize I need spaces here.

mlh758:
After using a lot of git tools, I ended up just sticking with the command line version. You would want
`git rebase -i @~5`

Git rebase lets you squash multiple commits into a single commit, change commit messages, and generally rewrite the history of a branch. The `@` symbol means `HEAD`, it's just a shorthand. In this case it means rebase the last 5 commits.

I can handle it when I merge your changes in though.

mlh758:
Looks like you got the formatting fixed. I merged your changes and then fixed a small code comment. I also rewrote the history a little bit. I squashed the format fix commits and stuff down into their previous commits just to keep the history a little cleaner. I also reworded the messages to get rid of the issue references.

PR #725, opened 2018-01-13
$variable naming & comments

SteveTheTechie:
Mostly a variable re-naming and commenting update.

1. Updated version number to 3.0.0 in code heading

2. Updated jQuery version dependency to 1.7+ due to deprecation of `$().delegate & $().bind()`, and use of `$().on()` instead

3. Detailed comments for each option to explain the permitted value types and the intended use of the option.

4. Switch default for selectedText option to be `# of # selected` to ensure that developers know that this form is supported.

5. Follow the popular convention of prefixing variable names of jQuery objects with a dollar sign for more clarity.

6. Use local variables for properties fetched more than once.

7. Ditch use of `$().bind()` in favor of `$().on()`, as bind is deprecated.

Unit tests have been updated.

PR #726, opened 2018-01-14
Implement selectedMax feature

SteveTheTechie:
The selectedMax option limits the number of selections to a maximum value.   If the user attempts to check an additional selection in the multi-select, a warning message is shown in the button for 1 second.

This feature is very useful for the "select up to a maximum of x out of y available selections" use case.  It implements a constraint approach.

An additional unit test has been also added for this feature.

**NOTE:**  This PR follows PR #725 in succession.

### DIFF
--- a/tests/unit/options.js
+++ b/tests/unit/options.js
@@ -1,82 +1,82 @@
 (function($){
 
-	module("options");
+   module("options");
 
-	test("noneSelectedText", function(){
-		expect(7);
-		var text;
+   test("noneSelectedText", function(){
+      expect(7);
+      var text;
 
-		el = $("select").multiselect({
-			noneSelectedText: 'None Selected'
-		});
+      el = $("select").multiselect({
+         noneSelectedText: 'None Selected'
+      });
 
-		// read from widget
-		text = el.multiselect("option", "noneSelectedText");
+      // read from widget
+      text = el.multiselect("option", "noneSelectedText");
 
-		equals( button().text(), text, 'on init, button reads "None Selected"');
-		el.multiselect("checkAll");
-		ok( button().text() !== text, 'after checkAll, button no longer reads "None Selected"');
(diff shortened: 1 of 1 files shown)


SteveTheTechie:
Will re-open later.

SteveTheTechie:
Re-opening.   Found the problem.  ;-)

SteveTheTechie:
There is a question of what to do with checkAll if a selectedMax constraint is in place.
Perhaps "Check All" cannot be shown or acted on if a constraint is in place.   
This also may present some challenges w/ Flip All. 

Perhaps any selections beyond the selectedMax limit are cleared?

BTW, I already use selectedMax (I call it maxSelected) constraint feature in my old code version that I am using in my web app.   

I am just trying to make sure I "tidy it up" here if other people are going to be using it, also.

mlh758:
Okay, I rebased your last branch to squash it down. That means the history is a little different now between these branches. Don't worry though, it's an easy fix! You might want to make a branch off this branch to test on the first time you do the `rebase --onto` command just in case.

`git rebase --onto <target_branch, probably ehynds/version3> HEAD~2`
Make sure your changes are still there and correct
`git push -f` <-- The force push let's you overwrite the remote branch with the new history

mlh758:
Sorry I didn't respond to your question sooner. I would say it would be best to disable select all when the selectedMax feature is enabled. I don't think it would make sense to trim the options list down, and just selecting the first x rows options wouldn't be helpful either.

SteveTheTechie:
Ok I will tweak this a bit more in a subsequent PR to address the concerns.

SteveTheTechie:
This code update is included in #728.  I am not sure if you want me to just close this and let you just merge 728?

mlh758:
Yeah if this is in 728 as well I'll just merge it all at once.

Issue #724, opened 2018-01-12
Opportunities for improved performance


SteveTheTechie:
Reference #717 comment 1.

Creating entire jQuery objects for what are essentially creation and manipulation of single DOM elements...

**Example 1:**   _(Ran inside a loop... lots of potential.)_
```
      var $item = $("<li/>").addClass(liClasses.join(' '));
      var $label = $("<label/>").attr({
        "for": inputID,
        "title": title
      }).addClass(labelClasses.join(' ')).appendTo($item);
      var $input = $("<input/>").attr({
        "name": "multiselect_" + id,
        "type": isMultiple ? "checkbox" : "radio",
        "value": value,
        "title": title,
        "id": inputID,
        "checked": isSelected ? "checked" : null,
        "aria-selected": isSelected ? "true" : null,
        "disabled": isDisabled ? "disabled" : null,
        "aria-disabled": isDisabled ? "true" : null
      }).data($(option).data()).appendTo($label);

      var $span = $("<span/>").text($(option).text());

```

**Example 2:**  _("this" will reference a single DOM element here... no need for jQuery)_

```
      this.header.on('click.multiselect', 'a', function(e) {
        var $this = $(this);
        if($this.hasClass('ui-multiselect-close')) {
          self.close();
        } else if($this.hasClass("ui-multiselect-all")) {
          self.checkAll();
        } else if($this.hasClass("ui-multiselect-none")) {
          self.uncheckAll();
        } else if($this.hasClass("ui-multiselect-flip")) {
          self.flipAll();
        }

```

SteveTheTechie:
I am digging into this a bit.   I have done some research and I found the following to be very informative.

[https://stackoverflow.com/questions/327047/what-is-the-most-efficient-way-to-create-html-elements-using-jquery](https://stackoverflow.com/questions/327047/what-is-the-most-efficient-way-to-create-html-elements-using-jquery)

[https://jsperf.com/jquery-element-creationyay/19](https://jsperf.com/jquery-element-creationyay/19)

mlh758:
Yeah, looking at the results in one of the replies `$( document.createElement('div') );` seems to be an order of magnitude faster. It also still gives us a jQuery object to play with to keep the code simple.

SteveTheTechie:
[...]

Issue #696, opened 2017-04-18
Add option to use .html() instead of .text() to get option display and set checkbox span display for decoration purposes


gatsy97d:
In line 197 of jquery.multiselect.js, when option isHtml is true instead of

$("<span/>").text($(option).text()).appendTo($label);

it will use

$("<span/>").html($(option).html()).appendTo($label);

mlh758:
You're looking for an additional option to allow HTML in the option tags?

gatsy97d:
Yes adding an option to use .html() instead of .text()
im still not familiar with contributing to github since I know where to update in the code.

SteveTheTechie:
This will be included in version 3.   A new option has been created.

Additionally, an option has also been created for the button text.   For example, I like to make my button list separate be '<br>' so that each selected list item shows on a separate line and the button grows vertically (to some limit as set by selectedList).

SteveTheTechie:
This issue can now be closed.   An option called htmlOptionText was added in the last pull request for version 3 that has been merged.

PR #728, opened 2018-01-17
Improve performance and fix issues

SteveTheTechie:
1. Add htmlButtonText option
2. Add htmlOptionText option - Addresses #696
3. Use more local variables instead of frequently referencing widget properties/options.
4. Use document.createElement() to more quickly create new DOM elements.  Addresses #717 comment 1 and #724.
5. Eliminate use of href="#" in header links.  Add cursor: pointer to jquery.multiselect.css for these links.
6. Move insertions to end of _create to limit browser reflowing lags.
7. Use more ternary operators to eliminate if/then's where it makes sense.
8. Create _updateCache() internal function... called from multiple places.
9. Ditch use of $.isFunction.  A simple typeof can be reliably used instead.
10. Add check in update() to reposition menu if button height has changed from saved value and the menu is open.  Addresses comment 5 in #717.
11. Save the current button height in position() to enable checking for button height changes that require menu re-positioning.  Addresses comment 5 in #717.

**Tests Updated**
12. Added test for namespace separation to core.js
13. Added selectedList:0 to tests in core.js to ensure that button text is correct for tests.
13. Fixed minWidth percentage test in options.js that was failing for non-Chrome browsers--this is likely due to floating point round off differences between browsers.

**Files updated:**
jquery.multiselect.js
jquery.multiselect.filter.js
jquery.multiselect.css
core.js
options.js

### DIFF
--- a/jquery.multiselect.css
+++ b/jquery.multiselect.css
@@ -7,7 +7,7 @@
 .ui-multiselect-header ul { font-size:0.9em }
 .ui-multiselect-header ul li { float:left; padding:0 10px 0 0; }
 .ui-multiselect-header a { text-decoration:none; }
-.ui-multiselect-header a:hover { text-decoration:underline; }
+.ui-multiselect-header a:hover { text-decoration:underline; cursor: pointer;}
 .ui-multiselect-header span.ui-icon { float:left; }
 .ui-multiselect-header .ui-multiselect-close { float:right; padding-right:0; text-align:right; }
 
--- a/src/jquery.multiselect.filter.js
+++ b/src/jquery.multiselect.filter.js
@@ -1,12 +1,11 @@
 /* jshint forin:true, noarg:true, noempty:true, eqeqeq:true, boss:true, undef:true, curly:true, browser:true, jquery:true */
 /*
- * jQuery MultiSelect UI Widget Filtering Plugin 3.0.0
+ * jQuery MultiSelect UI Widget Filtering Plugin 2.0.0
  * Copyright (c) 2012 Eric Hynds
  *
  * http://www.erichynds.com/jquery/jquery-ui-multiselect-widget/
  *
  * Depends:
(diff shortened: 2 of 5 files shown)


mlh758:
Thanks for fixing that test, I'm not sure when it started failing but I noticed it on your first PR. I poked at it a couple of times but hadn't had time to sit down and investigate it thoroughly.

SteveTheTechie:
When I looked into the test, it appears that the difference is always 1 or less.  You can get a difference of 1 if the parent div has an odd number of pixels width... This is why I thought that the floating point round off rationale is plausible.  I just fixed the test by making a difference of 1 or less passable.

mlh758:
I used to always see a difference of less than one which is why I added that rounding operation.

On my mac I was seeing a difference of about 12 - 20. The issue was coming up because the element's parent width was changing substantially between lines of code in the test.

I'll pull down your changes and see if it is still fixed. If it's not, your changes are still an improvement for other browsers and the onus isn't on you to fix specs that have decide to die on some browsers so I'll still merge it.

SteveTheTechie:
Getting accurate reliable widths is one area where jQuery has historically had a number of bugs, so it may well be that this is a jQuery related issue.  I would recommend keeping that in mind.  In some cases, it may be necessary to ditch jQuery because of particular features that have had issues in the past.

[https://www.google.com/search?q=jquery+width+bug&oq=jquery+width+bu&aqs=chrome.1.69i57j0l5.15910j0j7&sourceid=chrome&ie=UTF-8](https://www.google.com/search?q=jquery+width+bug&oq=jquery+width+bu&aqs=chrome.1.69i57j0l5.15910j0j7&sourceid=chrome&ie=UTF-8)

SteveTheTechie:
The extra commit is just minor stuff I missed.  Trying to keep the commits as close as possible to what I actually have in my editor.  ;-)

Issue #731, opened 2018-01-22
More Performance Tweaks++


SteveTheTechie:
I have previously developed some "no-jQuery", but "jQuery-like" DOM helper functions that allow doing efficient DOM operations in a sane way.  They are chainable functions which is why I say they are "jQuery-like".  I have adapted these for use in the widget and I have a version of _makeOption that is using these (it is faster).  They are below (at the bottom of this)...  Typical usage example:

```
var label = document.createElement('label')
	._setAttribute({ 'for': inputID,  'title': title})
	._addClass([isDisabled ? 'ui-state-disabled': null, isSelected && !isMultiple ? 'ui-state-active' : null, 'ui-corner-all'])
	._appendChild($input.get(0))
	._appendChild(span);

```
Do you have any objections to me using these to speed up the list creation?  Additionally, I want to get rid of the recursion in _buildOptions as it is preventing just doing one jQuery.append operation to add all list items from an array.  (one append from an array is faster)

```
	Element.prototype._setAttribute = function(attributes, value) {
		var elem = this;
		
		if (typeof attributes === "string") {
			elem.setAttribute(attributes, value);
			return elem;
		}		
		for (var name in attributes) {
			if (attributes[name] !== null)
				elem.setAttribute(name, attributes[name]);
		}
		return elem;
	};
	 
	Element.prototype._removeAttribute = function(attributes) {
		var elem = this;
		
		if (typeof attributes === "string") {
			elem.removeAttribute(attributes);
			return elem;
		}		
		var length = attributes.length;
		for(var x = 0; x < length; x++) {
			if (attributes[x])
				elem.removeAttribute(attributes[x]);
		}
		return elem;
	};
	
	Element.prototype._addClass = function(classes) {
		var elem = this;

		if (typeof classes === "string") 
			classes = classes.indexOf(' ') + 1 ? classes.split(' ') : [classes];		
		
		var length = classes.length;
		for(var x = 0; x < length; x++) {
			if (classes[x]) {
				if (elem.classList) 
					elem.classList.add(classes[x]);
				else 
					elem.className += ' ' + classes[x];
			}
		}
		return this;
	};
	
	Element.prototype._html = function(newhtml) {
		this.innerHTML = newhtml;
		return this;
	}
	
	Element.prototype._text = function(newtext) {
		this.textContent = newtext;
		return this;
	}
	
	Element.prototype._appendChild = function(childNode) {
		this.appendChild(childNode);
		return this;
[...]

Issue #730, opened 2018-01-22
iconSet option vs individual icon options


SteveTheTechie:
I would like to add the ability to use alternative icons for the widget.  However, even though I already added some individual options for specifying alternative icons, _it recently occurred to me that, in practice, a developer will likely just use all jQuery icons, all HTML Entities for icons, or all Font Awesome icons_--these are the alternatives I am accounting for currently.  (More could be added later.)

Right now, I have the following defined in my "not-yet PR'd" code:

```
	var msIcons = {
		'jQuery': {
			'open': '<span class="ui-icon ui-icon-triangle-1-s"></span',
			'close': '<span class="ui-icon ui-icon-circle-close"></span>',
			'checkAll': '<span class="ui-icon ui-icon-check"></span>',
			'uncheckAll': '<span class="ui-icon ui-icon-closethick"></span>',
			'flipAll': '<span class="ui-icon ui-icon-arrowrefresh-1-w"></span>'
		},
		'HTMLEntities': {
			'open': '<span class="html-entity">&#9660;</span>',
			'close': '<span class="html-entity" style="color:lightgrey;background-color:black;padding:0 2px;margin-right:1px;">&times;</span>',
			'checkAll': '<span class="html-entity">&check;</span>',
			'uncheckAll': '<span class="html-entity">&cross;</span>',
			'flipAll': '<span class="html-entity">&olarr;</span>'			
		},
		'FontAwesome': {
			'open': '<i class="fa fa-caret-down" aria-hidden="true"></i>',
			'close': '<i class="fa fa-times-circle" aria-hidden="true"></i>',
			'checkAll': '<i class="fa fa-check" aria-hidden="true"></i>',
			'uncheckAll': '<i class="fa fa-times" aria-hidden="true"></i>',
			'flipAll': '<i class="fa fa-undo" aria-hidden="true"></i>'
		}
	};

```
It seems like it may be better to just have an "iconSet" option to specify which predefined set of icons to use and then also provide a method to swap out an individual icon dynamically.  For example, maybe the open or close icons could be changed dynamically for indicating context.  However, to me, "dynamically" implies using a method rather than having an option.

Your thoughts?

SteveTheTechie:
Again, the reason for allowing for alternative icons is that the default jQuery icons are not scaleable and can look rather small on some screen resolutions.

mlh758:
[...]

Issue #729, opened 2018-01-21
showCheckAll vs checkAllText is null


SteveTheTechie:
It seems to me that having both showCheckAll and checkAllText options is superfluous, since you could just use a checkAllText === null option setting to imply that the related link should not be shown.  This also could allow for a checkAllText === '' (blank) setting to mean that the related icon, but not the text, should be shown.

This also applies to the showUncheckAll and the showFlipAll options (the latter being what I added).

I would like to ditch the showCheckAll, showUncheckAll, and showFlipAll options in favor of just using more values in the respective Text options to accomplish the same thing.

Your thoughts on this?

SteveTheTechie:
Yes

mlh758:
Your changes merged so closing this issue.

PR #732, opened 2018-01-23
Performance tweaks 2

SteveTheTechie:
Reference Issues:  #724 , #729, #730, #731

1. Added css/ folder and moved css files there.  Folder references updated.
2. Added iconSet option to be used for alternate icon sets.  Removed individual icon options.
3. Ditched showCheckAll, showUncheckAll, and showFlipAll.  Just set the respective text options to null to hide the respective links.
4.  Various performance tweaks:
-  Minimized number of appends.
-  Switched from $().appendTo() to $().append()--the latter is faster.
-  Reordered appends to minimize browser reflow lags and combine appends.
-  Eliminated recursion in _buildOptionList by implementing a secondary loop.
-  Implemented the use of $().toggle() more widely.


mlh758:
How did this end up comparing with the little speed test gist I provided?

package.json
```diff
@@ -10,13 +10,18 @@
...
   ],
   "repository": {
     "type": "git",
     "url": "https://github.com/ehynds/jquery-ui-multiselect-widget"
   },
   "main": "src/jquery.multiselect.js",
-  "dependencies": {},
+  "dependencies": {
+    "jquery": "^1.8.0",
+    "jquery-ui": "^1.8.0"
```
mlh758: I think it's actually jquery-ui 1.11.0 since I started using the `instance` method from the widget factory.

src/jquery.multiselect.filter.js
```diff
@@ -149,9 +149,9 @@
         });
 
         // trigger the change event on the select
-        if($inputs.length)
+        if(inputCount)
```
mlh758: This is referencing the variable from that for loop above, you should move the declaration to outside the for loop if you are going to use it elsewhere.

src/jquery.multiselect.js
```diff
@@ -26,48 +26,50 @@
...
+ * - https://jsperf.com/append-array-of-jquery-elements
+ * - https://gist.github.com/adrienne/5341713
  *
  */
 (function($, undefined) {
   // Counter used to prevent collisions
   var multiselectID = 0;
   var $doc = $(document);
 
+   var msIcons = {
```
mlh758: Maybe call this `defaultIcons` instead?

src/jquery.multiselect.filter.js
SteveTheTechie: inputCount is defined at the top of the function and used in two places, one of them being in the for loop and one of them being here.

package.json
SteveTheTechie: Will fix.

src/jquery.multiselect.js
SteveTheTechie: Will. change

SteveTheTechie:
I have not gotten to the gist yet.  I will look at that after I get this sorted out.   I am not very familiar with many of these online tools like gist and jsperf, so maybe I can get some more practice with them.  ;-)

I am not sure what happened w/ the tabs / spaces thing.  I thought I had addressed that.  ;-(

I am not familiar w/ the `?w=1` thing.

SteveTheTechie:
Some of the tabs were pre-existing.  If you pull up one of the demos from the master branch, for example, and turn on tab display in your editor, you can see that the existing files already have a number of tabs in them.  

No matter, though...  I went ahead and swapped out tabs for anything I touched.

mlh758:
It was just in the multiselect.js file that I was noticing it - it seemed like there were a bunch of white space only changes. I'm not as worried about formatting in the demo files.

A gist is just a link to some standalone code on GitHub. It's a nice way to keep and share little code samples.

You can add `?w=1` to the end of any GitHub diff URL and it will hide any white space only changes. Really nifty.

SteveTheTechie:
Added another commit to address the inputCount thing.   I had "file confusion".  lol

src/jquery.multiselect.filter.js
SteveTheTechie: **CORRECTION:** For some reason I did not realize you were referencing the jquery.multiselect.filter.js file here, even though GitHub clearly states that.  ;-D   Sorry... I changed this in another commit.

mlh758:
I plan on merging it today. I didn't have the chance to check out the changes last night though.

mlh758:
On my computer at work these changes bring the load time from 1.3 seconds on the speed test to about 1 second. 

That's certainly an improvement, and if there's a lot of widgets with a lot of items that would be a worthy change. Weirdly though, it looks like this when the page loads:
image


That doesn't happen on master. Did you see something similar? The other demos look okay.

Issue #727, opened 2018-01-14
Determining the best width


SteveTheTechie:
Reference issue #717, comments 2 and 8.

Here is the essence of how I automatically determine a "best width" that will accommodate all the select options...  Essentially a hidden clone of the menu is produced that is used for measuring.  The clone is set up specifically to allow the container div to assume the smallest width needed to accommodate the widest option.

Determining a single option height allows us to just multiply the number of options by the native select's size attribute to determine a menu height.

This is not perfect-there is room for improvement.  In particular, whether or not options can be wrapped should be set via a new option.  Wrapping affects the width and the appearance.

Also, the measuring can be a bit slow.   If we had a reliable way to quickly determine the widest option then we could just make a dummy menu clone with that single option for measuring which would be faster.  This is an approach that I not yet explored.

```
	// Added to allow automatic menu sizing,  8/17/2014, Steve James
	// Reference: http://stackoverflow.com/questions/1472303/jquery-get-width-of-element-when-not-visible-display-none
	// 		http://stackoverflow.com/questions/5903112/is-it-really-impossible-to-make-a-div-fit-its-size-to-its-content
	
	_measureMenu: function() {			
		// width: auto needed to reset width when dialogs 
                //    containing multi-select are re-opened.  
		// Display: inline-block needed to force width to best fit.
		
		// Make the hidden clone for measuring...
		var $clone = this.$menu.clone()
                                 .appendTo('body')
                                 .css({visibility:'hidden',width:'auto',display:'inline-block'})
                                 .find('ul')
                                 .css({display:'inline-block'})
                                 .end()
                                 .show();
				
		// Find outer height of a hidden menu item before we display menu.																											
		this.optionHeight = $clone.find('label').outerHeight();																																	
		
		// Find best width of hidden menu before we display it; 20 is for scroll bar		
		this.menuWidth = Math.max($clone.find('.ui-multiselect-header').width(),
                                         $clone.find('.ui-multiselect-checkboxes').width()) - 20;
		$clone.remove(); 
	},

```

SteveTheTechie:
I am currently digging into this again after several years.  Getting reliable measurements of DOM elements can be a real challenge at times.   Some of the relevant understandings that I have gained over time and via recent research:

[...]

Issue #569, opened 2014-08-06
Modifying multiselect list height in terms of rows, not pixels


smontanaro:
I'm not much of a JS/jQuery person, so I may be going about this all wrong, however...

There seems to be no obvious way to set the height of the <ul> widget generated by jquery.multiselect (again, consider that I know little about how all this JS/jQuery stuff works, so am probably missing something everyone else "just knows"). What little I was able to discover from looking at the code suggests that it never attaches units to the height either, so "px" are always assumed. This seems to be a bit of a regression in terms of abstraction. Plain old <select> widgets have a size attribute which is specified in rows. It would be nice if that size attribute was translated to something meaningful by this system. As an alternate, allowing setting of the height with attached units ("em"?) would be better than specifying them in pixels.

SteveTheTechie:
This will be addressed in my next PR.

SteveTheTechie:
Implemented by #735.   Closing this.

SteveTheTechie:
@mlh758 Implemented by #735. This can be closed.

PR #735, opened 2018-01-28
Peformance 3 and Dimensions

SteveTheTechie:
> Closes out #569 and #731--these can be closed.
> Partially addresses #717 and #727. (do not close yet)

1. Converted _makeOption to plain JS for speed.  This is the last specific jQuery-to-plain-JS conversion that I am planning.

2. Added addInputNames option to allow opting out (false value) of creating name properties for inputs.  
 These names sometimes duplicate what is already in use for the native select and just cause more inputs to be serialized and posted to the server than is needed. (slower)

3. Added caching of button width value, menu width value, menu height value, and positioned status and mechanisms to invalidate the cached status for each. The caching minimizes the recalculation of each of these values when the menu is successively opened.  (Just calculated on first open and re-used.)  

4. Re-wrote _setMenuHeight:
  - a. Implement a new "size" keyword for the menuHeight option that sets the menu height based on the number of rows specified in the native select's "size" attribute.  If "size" is specified, but no attribute is found on the native select, a default  size value of 4 rows is used like is the default for native multiple selects.
  - b. Also, _setMenuHeight is optimized to execute the least loop iterations to determine the appropriate menu height to use.
  - c. Lastly, the menu height is limited to the available window.innerHeight value.

5. Changed from using self.element.parent().outerWidth() to using self.element.parent().innerWidth() in _getMinWidth() to fix errors in % width calculations.

6. Tweaked the css file to remove unnecessary tag qualifications that degrade layout efficiency.  

7. Tweaked the css file to replace descendent selectors with child selectors where possible for improved layout efficiency.


CHANGELOG.MD
```diff
@@ -0,0 +1,13 @@
+Summary of Code Changes for jQuery UI Multiselect Widget, Version 3.0.0:
```
mlh758: Don't need a changelog file, we'll detail all the enhancements when we release the version 3 tag.

css/jquery.multiselect.css
```diff
@@ -1,25 +1,29 @@
+/*  References:
```
mlh758: I would prefer to stop including these reference comments. I was planning on going through and removing all of them from the old code once you were done merging changes.

I don't think they're really necessary and practices will change as web technologies evolve anyway.

src/jquery.multiselect.js
```diff
@@ -43,60 +47,66 @@
...
-      htmlButtonText: false,               // (true | false) If true, then the text used for the button's label is treated as html rather than plain text.
-      htmlOptionText: false,               // (true | false) If true, then the text for option label is treated as html rather than plain text.
-      disableInputsOnToggle: true,         // (true | false)
-      groupColumns: false                  // (true | false)
+   $.widget("ech.multiselect", {
+
+   // default options
+   options: {
+      header: true,                       // (true | false) If true, the header is shown.
+      height: 175,                        // (int | 'size') Sets the height of the menu in pixels.  If 'size' is instead specified, the native select's size attribute is instead for height.
```
mlh758: Your comment here is kind of confusing.

src/jquery.multiselect.js
```diff
@@ -43,60 +47,66 @@
...
       return elem;
     },
 
-    // Performs the initial creation of the widget
-    _create: function() {
-      var $element = this.element.hide();  // element property is a jQuery object
-      var elSelect = $element.get(0);      // This would be expected to be the underlying native select element.
+   // Performs the initial creation of the widget
+   _create: function() {
+      var $element = this.element.hide();                              // element property is a jQuery object per http://api.jqueryui.com/jQuery.widget/
```
mlh758: I don't really see a need for comments describing what a jQuery object is within a jQuery widget. I think you deleted them before too.

src/jquery.multiselect.js
```diff
@@ -43,60 +47,66 @@
...
+      var elSelect = $element.get(0);                                  // This would be expected to be the underlying native select element.
```
mlh758: It's expected to be... Is it not?

src/jquery.multiselect.js
```diff
@@ -113,66 +123,67 @@
...
-                  'title': elSelect.title, 
-                  'tabIndex': elSelect.tabIndex, 
-                  'id': elSelect.id ? elSelect.id  + '_ms' : null 
-                })
-           .prop('aria-haspopup', true)
-           .html('<span class="ui-multiselect-open">' + iconSet.open + '</span>');   // Necessary to simplify dynamically changing the open icon.
-
-        this.$buttonlabel = $( document.createElement('span') )
+            .addClass('ui-multiselect ui-widget ui-state-default ui-corner-all' + (classes ? ' ' + classes : ''))
+            .attr({ 'type': 'button', 'title': elSelect.title, 'tabIndex': elSelect.tabIndex, 'id': elSelect.id ? elSelect.id  + '_ms' : null })
```
mlh758: I would prefer you expanded it again, easier to read with the ternary hidden in there.

src/jquery.multiselect.js
```diff
@@ -184,105 +195,125 @@
...
+                          "checked": isSelected ? "checked" : null,
+                          "aria-selected": isSelected ? "true" : null,
+                          "disabled": isDisabled ? "disabled" : null,
+                          "aria-disabled": isDisabled ? "true" : null
+                         };
+      for (var name in inputAttribs) {
+         if (inputAttribs[name] !== null)
+            input.setAttribute(name,inputAttribs[name]);
+      }
+      if ('dataset' in option) {
```
mlh758: The dataset attribute isn't supported in IE 10, it only got added in 11 and for the time being I would prefer to maintain IE 10 support.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

src/jquery.multiselect.js
```diff
@@ -579,71 +623,116 @@
...
+            $menu = self.$menu;
+            var appendEl = self._getAppendEl();
+            var cssVisibility = $menu.css('visibility');
+            var cssWidth = $menu.css('width');
+            var cssDisplay = $menu.css('display');
+            var cssulDisplay = $menu.find('ul:eq(0)').css('display');
+            $menu.appendTo('body').css({visibility:'hidden', width:'auto', display:'inline'}).find('ul').css('display', 'inline');
+            var autoWidth = $menu.width();
+            $menu.css({visibility:cssVisibility, width:cssWidth, display:cssDisplay}).find('ul').css('display', cssulDisplay).appendTo( appendEl );
+            console.log('AUTO WIDTH: ', cssWidth, autoWidth);
```
mlh758: Delete the console statement.

src/jquery.multiselect.js
```diff
@@ -579,71 +623,116 @@
...
+    _setMenuWidth: function(recalc) {
+      var self = this;
+      if (self._savedMenuWidth && !recalc)
+         return;
+
+      self._positioned = false;             // Force menu positioning to be adjusted on next open.
+      var width = self.options.menuWidth;
+
+      if (!width) {                                                   // Exact width not provided; determine appropriate width.
+         width = self._savedButtonWidth || self.$button.outerWidth();
```
mlh758: You're in the recalc section here, and still using the saved value. Is that intentional?

mlh758:
You're killing me with these whitespace changes...

Was it necessary to switch from `this` to `self` everywhere like that?

CHANGELOG.MD
SteveTheTechie: Ok.

css/jquery.multiselect.css
SteveTheTechie: Ok

src/jquery.multiselect.js
SteveTheTechie: It is referencing a button width in the setMenuWidth function.  Since the button width was just likely already calculated, this just tries to use the recently recalculated cached value.

src/jquery.multiselect.js
SteveTheTechie: Oops!  Sorry about that.  Will remove.  

Also, I goofed as I did not intend to include this auto width functionality in this.   I am still fiddling with it.  I will remove it, also.  (sorry)

SteveTheTechie:
My apologies.   I apparently was a bit hasty in posting this, and I overlooked some things, and inadvertantly included some things I did not intend to include.

Also I will revert some of the things you mention.

src/jquery.multiselect.js
SteveTheTechie: Michael,  if you want other people to contribute, then you need to make sure that other people can quickly and easily understand the code and the foundations it is based on.   If you are really committed to getting contributions from others, then it is unwise and insufficient for you to be the only one who really understands the code,  Also, useful "understanding" is often more than a literal understanding, sometimes an understanding of the context in which the code was created is also helpful.  This is why I find documenting references in comments to be helpful.

Even the much vaunted jQuery's source code has a lot of comments in it.  They also reference tickets and other websites in their code.  Again, it ensures that the code can be easily picked up and updated by people other than the current maintainers. 

Frankly, what does it matter, anyway?  If the code is minified (uglified) as it should be, then all the comments will be removed anyway from the actual running code.

src/jquery.multiselect.js
SteveTheTechie: Will do.

src/jquery.multiselect.js
SteveTheTechie: Thanks.   Let me see if I can find a work around.

src/jquery.multiselect.js
SteveTheTechie: The 'size' setting just makes the height of the multiselect accommodate what is needed for the # of rows specified via the size setting of the native select.   This is actually very useful... it is something I use a lot w/ my implemented version of my old fork in my web app.

I am totally open to less confusing wording.

SteveTheTechie:
I think this should be ready for you to take another look.  Thank you for your patience with me.

src/jquery.multiselect.js
mlh758: It's mostly a grammar thing:
> the native select's size attribute is instead for height.
could maybe be
> the native select's size in rows attribute is used instead for height

src/jquery.multiselect.js
mlh758: Ah, I see. That makes sense.

mlh758:
I'll try to get this code reviewed again today and test it out.

I saw this comment in my e-mail, although it doesn't appear to be on this chain anymore. It's worth discussing so I'm going to quote it in and reply anyway:

> Michael, if you want other people to contribute, then you need to make sure that other people can quickly and easily understand the code and the foundations it is based on. If you are really committed to getting contributions from others, then it is unwise and insufficient for you to be the only one who really understands the code, Also, useful "understanding" is often more than a literal understanding, sometimes an understanding of the context in which the code was created is also helpful. This is why I find documenting references in comments to be helpful.

Useful code comments document document parameters, side effects, and purpose when it isn't clear. Your comment [here](https://github.com/ehynds/jquery-ui-multiselect-widget/pull/735/files?w=1#diff-deab69d9cf58b93be9077b382654c121R1010) is an example of a good comment. At first glance, it wouldn't necessarily be obvious to someone new why there are two of these. Reference comments require a contributor to follow a link, that may eventually be dead, and may no longer really apply, rather than actually explaining what the code below did that was so unusual that it required a comment instead of being clear on it's own what the purpose was.

> Even the much vaunted jQuery's source code has a lot of comments in it. They also reference tickets and other websites in their code. Again, it ensures that the code can be easily picked up and updated by people other than the current maintainers.

It's not comments I take issue with, I'd love for this source code to eventually adhere to the [Google style guide](https://google.github.io/styleguide/jsguide.html#jsdoc) for documentation. There are also a couple of places in this source code where jQuery issues are referenced. Those are useful pieces of information and I have no intention of deleting them.

> Frankly, what does it matter, anyway? If the code is minified (uglified) as it should be, then all the comments will be removed anyway from the actual running code.

I would prefer the comments that do need to be in place directly, and in plain language, explain the purpose (or side effects) of the code they are describing. Excessive comments can be just as detrimental to understanding as a lack of comments. Especially if they're used as a band-aid over confusing code that could be made less confusing.

mlh758:
> I think this should be ready for you to take another look. Thank you for your patience with me.

You're contributing a lot of really helpful changes to the widget, so you're not testing my patience at all. I expect there to be some changes changes necessary on any code review that big.

mlh758:
Looks like this reduced the load time on the speed test from ~1 second on my work computer to .3 seconds. Impressive.

PR #741, opened 2018-02-02
Fix #740 Group label inside an li tag

### LINKED ISSUE #740: Generated HTML markup for Option Groups is invalid
Reference [https://validator.w3.org/#validate_by_input](https://validator.w3.org/#validate_by_input)

The following models the way that anchor tags are used in the generated markup for option groups. However, this markup is flagged as invalid HTML by the HTML validator.  (I suspected it would be.)
```
<!DOCTYPE HTML SYSTEM>
<html>
<head>
<title>tester</title>
</head>
<body>
<ul>
<a href="https://www.google.com/">test</a>
<li>test2</li>
</ul>
</body>
</html>
```

If the anchor can somehow be wrapped in a `<li>`, it is then valid... e.g. the following validates fine:
```
<!DOCTYPE HTML SYSTEM>
<html>
<head>
<title>tester</title>
</head>
<body>
<ul>
<li><a href="https://www.google.com/">test</a></li>
<li>test2</li>
</ul>
</body>
</html>
```

mlh758:


mlh758:
@SteveTheTechie

css/jquery.multiselect.css
```diff
@@ -19,7 +19,7 @@
 .ui-multiselect-checkboxes li { clear:both; font-size:0.9em; list-style: none; padding-right:3px; }
 .ui-multiselect-checkboxes > .ui-multiselect-optgroup { padding: 3px; }
 .ui-multiselect-columns { display: inline-block; vertical-align: top; }
-.ui-multiselect-checkboxes > .ui-multiselect-optgroup > a { border-bottom:1px solid; cursor: pointer; display:block; font-weight:bold; margin:1px 0; padding:3px; text-align:center; text-decoration:none; }
+.ui-multiselect-checkboxes > .ui-multiselect-optgroup > li > a { border-bottom:1px solid; cursor: pointer; display:block; font-weight:bold; margin:1px 0; padding:3px; text-align:center; text-decoration:none; }
```
SteveTheTechie: Per [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul), if the list is properly nested. the optgroup class name would be on an li in the top level list, and the anchor tag would also land in that li tag.  Therefore, I think if the list is properly nested, then no change is needed here.

src/jquery.multiselect.filter.js
```diff
@@ -181,7 +181,7 @@
       // show/hide optgroups
       $groups.each(function() {
         var $this = $(this);
-        if (!$this.children('li').filter(':visible').length)
+        if (!$this.children(':not(:first-child)').filter(':visible').length)
```
SteveTheTechie: For proper nesting, I think you need to find the li's that are descendants of the current li containing the anchor.

css/jquery.multiselect.css
mlh758: The anchor tag is under an `li` which is the only valid descendant of a `ul`. This fixes the issue of an `a` tag being nested under a `ul`.

What are you hoping to have changed?

src/jquery.multiselect.filter.js
mlh758: The [accepted](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li) children of an `li` is any [flow content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Flow_content)

mlh758:
I see what you're saying, there's an additional issue where the option groups also need to be contained in an li.

SteveTheTechie:
I noticed the other issue later... sorry for my oversight.

mlh758:
No problem, just confusing.

mlh758:
Okay, updated the commit. The option group `ul` is now nested under a parent `li`. The `a` tag is under that `li` as well.

```
<ul class= "ui-multiselect-checkboxes ...">
  <li class="ui-multiselect-optgroup">
    <a>label</a>
    <ul>
      <li />
      <li />
    </ul>
  </li>
</ul>
```

mlh758:
@SteveTheTechie Now that it's updated does this look good to you?

src/jquery.multiselect.js
```diff
@@ -307,13 +307,16 @@
...
-
-          list.push($optionGroup);
+          var $optGroupItem = $( document.createElement('li') )
+            .addClass('ui-multiselect-optgroup' 
+            + (self.options.groupColumns ? ' ui-multiselect-columns' : '') 
+            + (elem.className && ' ') + elem.className)
+            .append( $( document.createElement('a') )                  
+              .text( elem.getAttribute('label') ) );
+          var $optionGroup = $( document.createElement('ul') ).append(options)
+          $optGroupItem.append($optionGroup)
```
SteveTheTechie: A nitpick:  I would suggest breaking out the anchor to save an append call:

var $optGroupItem = $( document.createElement('li') )
           .addClass('ui-multiselect-optgroup' 
             + (self.options.groupColumns ? ' ui-multiselect-columns' : '') 
              + (elem.className && ' ') + elem.className);
var $optGroupAnchor =$( document.createElement('a') ).text( elem.getAttribute('label') );
var $optionGroup = $( document.createElement('ul') ).append(options);

$optGroupItem.append($optGroupAnchor, $optionGroup);

src/jquery.multiselect.js
```diff
@@ -474,7 +477,7 @@
         e.preventDefault();
 
         var $this = $(this);
-        var $inputs = $this.parent().find('input').filter(':visible:not(:disabled)');
+        var $inputs = $this.parent().find('ul input').filter(':visible:not(:disabled)');
```
SteveTheTechie: I would just try the following since the ul will be the next sibling of the anchor (no need to find the parent):

`var $inputs = $this.next('ul').find('input').filter(':visible:not(:disabled)');`

src/jquery.multiselect.filter.js
```diff
@@ -181,7 +181,7 @@
       // show/hide optgroups
       $groups.each(function() {
         var $this = $(this);
-        if (!$this.children('li').filter(':visible').length)
+        if (!$this.children('ul').children('li').filter(':visible').length)
```
SteveTheTechie: Maybe just use `(!$this.find('li').filter(':visible').length)` since "find" finds descendants vs direct children as "children" does?

mlh758: I was thinking it would be nice to be explicit about the structure I was traversing, but it seems `find` is also generally faster so I'll just go with that.

mlh758:
Made the suggested changes and merged.

SteveTheTechie:
You ditched the css file update in your updated commit and now the display is off.

mlh758:
I have the version3 branch checked out on my local and it looks normal with and without option groups. What are you seeing?

SteveTheTechie:
Ok, maybe I am missing something.  Let me take another look.

mlh758:
I removed the css file change in the first round of revisions from your suggestion here:

>Per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul, if the list is properly nested. the optgroup class name would be on an li in the top level list, and the anchor tag would also land in that li tag. Therefore, I think if the list is properly nested, then no change is needed here.

You were correct, once it was nested correctly I didn't need that change.

SteveTheTechie:
Ok, well maybe I mis-typed or goofed... I apologize.  The change created another couple of impacts that we are not handled properly by the CSS.   Previously, the optgroup class was on the `<ul>` container, but now it is on a `<li>`.  So, we have `<li>` elements that have optgroup labels and contain the nested optgroups and we have plain "regular" `<li>` elements (the "leaves").  The `<ul>` container is now inside the optgroup `<li>` element.

Don't get me wrong... I agree with what you did--the HTML is proper and validateable.  I just think we need to also adjust the CSS to account for the change in markup.

**OLD HTML Structure:**
```
ul
   ul
      a
      li
      li
      (etc.)
   /ul
  (etc.)
/ul
```
**OLD Look:**
capture-multiselect-old

**NEW HTML Structure:**
```
ul
   li
      a
      ul
         li
         li
         (etc.)
      /ul
   /li
  (etc.)
/ul
```
**NEW Look: (notice extra padding and smaller font)**
capture-multiselect-new
capture-multiselect-new2

> Bottom Line:
>---
> Since we now have two types of `<li>` elements in the new markup scheme, but only one general CSS rule for `<li>`'s, I think what is happening is that we are seeing the CSS rule being applied twice cumulatively, once for the optgroup `<li>`'s and once for the nested `<li>`'s.  Also, the rule for the optgroups that applies padding should really go on the nested `<ul>`'s.
> 
> I have figured out a fix that only needs two CSS rules to be adjusted.  Unfortunately, we have to be able to target the list items that are not nested optgroups... a bit messier.  Need to use `:not(.ui-multiselect-optgroup)` to target those.  ☹️

SteveTheTechie:
**Changing:**
```
.ui-multiselect-checkboxes li { clear:both; font-size:0.9em; list-style: none; padding-right:3px;}
.ui-multiselect-checkboxes > .ui-multiselect-optgroup { padding: 3px; } 
```
**... to:**
```
.ui-multiselect-checkboxes li:not(.ui-multiselect-optgroup) { clear:both; font-size:0.9em; list-style: none; padding-right:3px;}
.ui-multiselect-checkboxes > .ui-multiselect-optgroup > ul { padding: 3px; } 
```
**... seems to fix the CSS.**

mlh758:
I reverted the commit for now now. I had noticed that on my first round of changes and fixed it and then apparently re-introduced it.

PR #742, opened 2018-02-05
Updating to QUnit 2.5, some spec formatting

mlh758:
@SteveTheTechie You mentioned getting QUnit up to date, I took care of that here.

Unfortunately, there was also some formatting to fix so the diffs are a bit of a mess. The main takeaways are that most QUnit operations (`module`, `ok`, `equals`) now need to have either QUnit or assert prefixed to them. 

* `assert` is passed as a parameter to each test function.

* `asyncTest` is gone, and instead `done` is declared using `assert.async()`

* There were some globals declared in core.js and used in the other files. I moved them to setup.js

* `setup` and `teardown` on module are gone in favor of things like `beforeEach` and `afterEach`


SteveTheTechie:
So mainly just updating function calls, then. 😃 

If I could make a suggestion...  There really is no reason to keep qunit.js and qunit.css in the repo.  Those can be pulled in from the jQuery CDN via the index.html file:
https://code.jquery.com/qunit/   (specific versions can be referenced)

I would suggest just updating the index.html file and deleting qunit.js and qunit.css from the repo.
A couple less files to deal with.  😉

mlh758:
Yup, it was pretty straightforward. A couple things got renamed or deprecated and the behavior of async changed a tiny bit.

SteveTheTechie:
I downloaded it and am working w/ it now.  Went from 9 failing tests in prior version to 29 failing tests in this version.

mlh758:
Could be from merge conflicts or something. There aren't any failing on the branch itself.

SteveTheTechie:
Nope... my fault.   I changed something between the two times running the tests.  😬 

Anyway, there is something in the tests that I think should be updated.  I have noticed that many of them tend to rely on a presumed set of default options in the code.  This sometimes causes test errors when changing options for the sake of testing.  It would be better to set a complete options object specific to  the tests that is tweaked for each test.   This could maybe be done by creating a helper function called _opts() (or whatever) that contains the complete set of default options that the tests are based on and if the function takes an argument, it could apply it and return the complete object.

Example usage:  `$("#myselect").multiselect( _opts({height: '2in'}) );`

This would help ensure that tests are ran with a consistent set of default options so that there are not spurious test failures.

SteveTheTechie:
> The common set of options is what is defined in the multiselect.js file, and the constructor does the logic of specifically overriding them for us. 

That is my point.  The overrides are not always complete in the tests.  Also, some tests depend on a specific default option value that I do not think should be presumed.   If people cannot successfully run the unit tests if they change the default option values in the code, I see that as problematic, since that really tests whether the options have changed not whether the code runs correctly.

SteveTheTechie:
In any case, this PR looks fine to me.   Could you go ahead and merge it.?

I want to post a PR that depends on these updated tests and would like to just fork off of the version3 branch.

mlh758:
> If people cannot successfully run the unit tests if they change the default option values in the code, I see that as problematic

The unit tests are defining behavior. If the behavior has changed, a unit test should need to be updated. If a default option has changed, that means the default behavior of the widget, likely what most people will be using, has been changed. I would expect to have to update tests to accommodate that.

Do you have an example of where this is being done in a way that you consider unreasonable?

Issue #737, opened 2018-01-29
Placeholder text and maxselected event.


SteveTheTechie:
Chosen is another one of these multiselects (other than Select2) that I am aware of:
[https://harvesthq.github.io/chosen/](https://harvesthq.github.io/chosen/)

A couple of interesting options it has are support for using the native select's placeholder text and it also allows for a maxselected event handler to be defined.

SteveTheTechie:
Closed by #747.

PR #744, opened 2018-02-06
Dimensions 2

SteveTheTechie:
Addresses #717 (partially) , #727, #737 (partially)

- [x] 1. Updated sizing/dimensions code.
- [x] 2. Added text wrapping option.
- [x] 3. Using getBoundingClientRect()
- [x] 4. Ditch globals
- [x] 5. Update tests.
- [x] 6. Update CSS
- [x] 7. Add support for placeholder text.  (via added commit)

_Note that this includes the Qunit / tests update by @mlh758_


SteveTheTechie:
@mlh758  When you have had a chance to look at this, there are a few things I would to discuss related to defaults and default behavior for some of the new options.

mlh758:
Do you want to address those questions before merging this PR?

tests/unit/options.js
```diff
@@ -194,11 +194,15 @@
...
 
-      // change height and re-test
+      // change width and re-test
       width = 300;
       el.multiselect("option", "menuWidth", width).multiselect('refresh');
       assert.equal(menu().parent().find(".ui-multiselect-menu").outerWidth(), width, 'changing value through api to ' + width);
 
+      width = "3in";
+      el.multiselect("option", "menuWidth", width).multiselect('refresh');
+      assert.equal(menu().parent().find(".ui-multiselect-menu").outerWidth(), 3 * 96.0, 'menuWidth supports strings suffixed with various units as well as integer px values');
```
mlh758: Can you update the assert message here with the `in` unit?

tests/unit/options.js
```diff
@@ -89,7 +89,7 @@
           selectedMax: 2
       });
 
-      checkboxes = el.multiselect("widget").find(":checkbox");
+      var checkboxes = el.multiselect("widget").find(":checkbox");
```
mlh758: Good catch adding all these vars.

src/jquery.multiselect.js
```diff
@@ -118,18 +123,21 @@
...
+              'title': elSelect.title,
+              'tabIndex': elSelect.tabIndex,
+              'id': elSelect.id ? elSelect.id  + '_ms' : null
             })
             .prop('aria-haspopup', true)
             .html('<span class="ui-multiselect-open">' + iconSet.open + '</span>');    // Necessary to simplify dynamically changing the open icon.
 
       this.$buttonlabel = $( document.createElement('span') )
-            .html(options.noneSelectedText)
+            .html(options.noneSelectedText || $element[0].placeholder || 'Select options')
```
mlh758: Do you need the third item in the `||` clauses? The description made it sound like if the option was null it would use the placeholder. If it's null and there's no placeholder I think I would prefer it do nothing or log a warning instead of having two places for the nonSelectedText option in the code.

src/jquery.multiselect.js
```diff
@@ -284,13 +292,13 @@
...
-     * This clears the items currently in $checkboxContainer
+     * This clears the items currently in $checkboxes
      * Defers to _makeOption to actually build the options
      * Resets the input ID counter
      * @param {object} $element element whose option/group tags need to be converted
-     * @param {object} $checkboxContainer widget's container to append the built options to
+     * @param {object} $checkboxes widget's container to append the built options to
      */
-    _buildOptionList: function($element, $checkboxContainer) {
+    _buildOptionList: function($element, $checkboxes) {
```
mlh758: Now that the recursion is gone, we could probably use the instance variables of the widget instead of passing these parameters in. Don't need to fix it in this PR, just something I wanted to make a note of.

src/jquery.multiselect.js
```diff
@@ -398,6 +409,9 @@
 
       self._setButtonValue(value, isDefault);
 
+      if ( !/\bbutton\b/.test( options.wrapText ) )
```
mlh758: For consistency I prefer that even single line if statements be wrapped in braces.

src/jquery.multiselect.js
```diff
@@ -726,11 +763,24 @@
...
+      // this._selectWidth set in _create() for native select element before hiding it.
+      var width = this._selectWidth || this._getBCRWidth( this.element );
+      var buttonWidth = this.options.buttonWidth || '';
+      if (/\d/.test(buttonWidth)) {
+         var parsed = this._parse2px(buttonWidth, this.element);
+         var pixels = parsed.px;
+         var minimax = parsed.minimax;
+         width = minimax < 0 ? Math.max(width, pixels) : ( minimax > 0 ? Math.min(width, pixels) : pixels );
+      }
+      else // keywords
```
mlh758: Same thing with the braces on all these.

mlh758: Keywords?

src/jquery.multiselect.js
```diff
@@ -865,6 +1013,8 @@
     /**
      * Toggles the checked state on options within the menu
      * Potentially scoped down to visible elements from filteredInputs
+     * Groups don't show up in the filter search right now, so both
```
mlh758: I deleted this portion of the comment intentionally, it wasn't true.

src/jquery.multiselect.js
```diff
@@ -976,7 +1126,10 @@
 
       // show the menu, maybe with a speed/effect combo
       // if there's an effect, assume jQuery UI is in use
-      $.fn.show.apply($menu, effect ? [ effect, speed ] : []);
+      if (effect)
```
mlh758: Same thing with braces.

mlh758:
There seems to be an issue around `refresh` now. If you add an item after the initial instantiation of the widget, the menu doesn't resize or wrap tax to accommodate.

Before:
image


After:
image

mlh758:
I created a [dialog demo](https://gist.github.com/mlh758/b1b24931db7705441f9786bd68cf3569#file-dialog-htm) in the same gist I had the speed test demo in. It looks like the width doesn't get calculated when the widget is in a dialog. Prior to this it would just wrap the text.

src/jquery.multiselect.js
SteveTheTechie: That is fine.  I can take that out.

src/jquery.multiselect.js
SteveTheTechie: Agreed.

src/jquery.multiselect.js
SteveTheTechie: ok

src/jquery.multiselect.js
SteveTheTechie: Keywords :  e.g.  'auto'    The comment is just emphasizing that keyword options drop to the else.

src/jquery.multiselect.js
SteveTheTechie: Oh, did I revert the delete by accident?

tests/unit/options.js
SteveTheTechie: Actually, this is a result of you updating Qunit.  👍 

It has a new option you can set when running it to find all the global variables and flag them as test failures.

tests/unit/options.js
SteveTheTechie: ok

SteveTheTechie:
Look at the wrapText option...   It should be set as 'button,header,menu' as the default.  It looks like I goofed and forgot to set that before committing.

SteveTheTechie:
@mlh758  Ok, made changes.   

Can you please take another look at the layout issues you were seeing?  I am hoping it was just a case of me previously forgetting to change the default value for an option back before committing.

SteveTheTechie:
As to what I wanted to discuss...  

- Height setting...

1.  The height of the menu is either a specific value (w/ any of various units), 'size', or 'auto'.   Size uses the native select size attribute to match the number of displayed rows.  Auto attempts to maximize the height to show all items up to the limit of the available window height.  The available window height is the constraint for *all* height settings--cannot set a height that exceeds that.  'Auto' is essentially the implicit fall-back in the code.

- Could you please take another look at the Select2 code I excerpted in #727 related to auto width?
- Width settings...

1.  The width of the native select is captured in _create before hiding it.
2.  Right now the general logic is that the width of the native select is used as a fall back for the button width, and the button width is used as a fall back for the menu width.
3.  Thus, if you set buttonWidth = null and menuWidth = null, then both should get set to the same width as the native select.
4.  If you set ">275px" for the button, it is treated as a minimum width for the button, meaning that if the select element width exceeds that, it is used--otherwise, it is 275px.
-  However, no min-width is set in the CSS like Select2.  An alternative would be to just set the CSS min-width to 275px in this case and not set a hard width, but I am not sure it matters since the widget does not support dynamically resizing it, except by methods which can be coded to call for a resize.
-  Also, what 'auto" _should_ mean for the button width is a bit uncertain for me.  If you kill the text wrapping for the button, it can grow in width, but whether that is practical is up for debate.  There is no intrinsic width check for the button since it is native button element... just set width: auto for it.
4.  If you set ">275px" for the menu, it is treated as a minimum width for the menu, meaning that if the button element width exceeds that, it is used--otherwise, it is 275px.
-  However, again, no min-width is set in the CSS like Select2.  An alternative would be to just set the CSS min-width to 275px in this case and not set a hard width, but I am not sure it matters since the widget does not support dynamically resizing it, except by methods which can be coded to call for a resize.  If you are going to add an option, than the add option method is supposed to invalidate cached values, which causes the them to get recalc'd again the next open.
-  If you set the menuWidth to 'auto', then the intrinsic / "shrink-wrapped" width of the header and checkbox container is determined and used to set a width for the menu.  However, again, this is not set as a min-width like Select2 does, it is set as a hard width.  Also, this width depends on whether text wrapping is on or off. (Typically, I think you would want it off for this case.)
- If the menu width is set differently than the button via an explicit menuWidth option setting (vs null), there is currently nothing that goes back to force the button width to match it.  If using groupColumns, the probably makes sense, but there is a question of whether some mechanism should exist to force the code to go back and change the button width after setting the menu width.   I think in most cases, you would just leave the menuWidth option as null to force a matched width, but for the case of menuWidth: 'auto', you could certainly end up with width that is wider than the button with no easy mechanism to fix it.  It did occur to me that maybe buttonWidth: 'auto' & menuWidth: 'auto' would force the menu width code to go back and fix the button width to match.  (but I am on the fence w/ that)

mlh758:
Checked it out again, layout issues are fixed and that 100% issue I showed on the base branch is also fixed.

mlh758:
>However, no min-width is set in the CSS like Select2. An alternative would be to just set the CSS min-width to 275px in this case and not set a hard width, but I am not sure it matters since the widget does not support dynamically resizing it, except by methods which can be coded to call for a resize. Also, I do not think that min-width is universally supported.

min-width has good browser support for the basic pixel/percentage values. But yes, we only really resize at specifically coded moments, so I'm not really sure if that would give us any extra benefit. I think it might be worth adding a wiki example that shows people how to hook into `window.resize` and have the widget adjust its own dimensions.


> Also, what 'auto" should mean for the button width is a bit uncertain for me. If you kill the text wrapping for the button, it can grow in width, but whether that is practical is up for debate. There is no intrinsic width check for the button since it is native button element... just set width: auto for it.

I think having the text wrapping on by default is the best way to go. It does seem to grow in auto mode now which, if someone wants that, is functional. I'm not sure on the practicality of it either but if it works and isn't the default behavior then I'm okay with it.

> If the menu width is set differently than the button via an explicit menuWidth option setting (vs null), there is currently nothing that goes back to force the button width to match it. 

To me that seems like the expected behavior of going out of your way to set the widths differently. I think the user could call `multiselect({menuWidth: null})` to get it to resize after instantiation anyway. I would need to test that though to make sure.

> It did occur to me that maybe buttonWidth: 'auto' & menuWidth: 'auto' would force the menu width code to go back and fix the button width to match. (but I am on the fence w/ that)

I don't think that's necessary since the default behavior is matching widths. If the user is changing these options it is presumably to no longer have them match.

> I tend to think that Select2 is trying to have a responsive widget that automatically adjusts its dimensions to fit its parent container. The question is whether we should be striving for that. I tend to think not, but I thought I would pose the question.

I agree with you here. At some point, if too much of our behavior mimics theirs, it might be worth asking if it's really worth having a separate widget from theirs. It's probably our differences in behavior that will keep this project valuable, as long as those differences aren't degrading the user experience.

> Could you please take another look at the Select2 code I excerpted in #727 related to auto width?

What is the takeaway I should be getting from that snippet? Their use of the css property?

SteveTheTechie:
> > Could you please take another look at the Select2 code I excerpted in #727 related to auto width?

> What is the takeaway I should be getting from that snippet? Their use of the css property?

Yes.  I was just referencing that in my comments about min-width and whether I should maybe be setting it or any other CSS.

No matter... I think I have my answer from your other comments.  😉

Issue #733, opened 2018-01-23
Ditch _toggleChecked replacement in jquery.multiselect.filter.js


SteveTheTechie:
The redefinition of _toggleChecked in jquery.multiselect.filter.js seems like overkill, particularly since the difference from the default definition is small and the behavior difference could just be handled by some sort of flag that the filter code sets on the main instance to cause the difference in logic.  Right now, you essentially have to keep the two definitions in sync if any changes to them are made.

My thinking is that the flag could either be set via a new call back that the filter code calls on the instance, or a new "filtered" parameter could be added to _toggleChecked that the filter code makes use of via "currying" the function.  (baking in the parameter value when filtering is in use)

Your thoughts?

mlh758:
Yeah if we can get rid of it I would be all for it. I haven't looked at that portion of the code much yet. It's something I can investigate. I'm pretty fond of function programming concepts but I don't have much of an opinion yet since I haven't looked at that code much.

SteveTheTechie:
I appears that when we changed to valid HTML markup we broke the select by optgroup functionality for the filtered version.   I am looking at that, and will probably just include the fix with the PR for this issue.

mlh758:
> I appears that when we changed to valid HTML markup we broke the select by optgroup functionality for the filtered version. I am looking at that, and will probably just include the fix with the PR for this issue.

What part of it broke? That was working when I merged the PR in, both filtered and unfiltered. Clicking by option group title and filtered and clicking the check all / uncheck all buttons.

SteveTheTechie:
Sorry... I jumped the gun.  Hence, I deleted the comment.   The sky is not falling.  lol  😆

mlh758:
Oh, woops. I didn't notice it was deleted last night. I just see the e-mails and go. I'm just gunna... delete all these comments and we can pretend it never happened.

SteveTheTechie:
Closed by merge of #745

PR #745, opened 2018-02-08
Eliminate replacement _toggleChecked code in multiselect.filter.js

SteveTheTechie:
Handles #733 

Replaces the separate _toggleChecked function code in multiselect.filter.js with instead using a parameter that changes the context for _toggleChecked in multiselect.js .

Also, a few housekeeping updates.

_Note: The events.js file that I pulled has tabs in it. What I am pushing back has spaces instead. (consider this fair warning...)

### DIFF
--- a/src/jquery.multiselect.filter.js
+++ b/src/jquery.multiselect.filter.js
@@ -122,7 +122,7 @@
       // is fired, only the currently displayed filtered inputs are checked
       var $instance = this.instance;
       $instance._oldToggleChecked = $instance._toggleChecked;
-      $instance._toggleChecked = function(flag, group, filteredInputs) {
+      $instance._toggleChecked = function(flag, group) {
          $instance._oldToggleChecked(flag, group, true);
       };
 
--- a/src/jquery.multiselect.filter.js
+++ b/src/jquery.multiselect.filter.js
@@ -71,11 +71,11 @@
           else if(e.which === 27) {
             $element.multiselect('close');
             e.preventDefault();
-          } 
+          }
           else if(e.which === 9 && e.shiftKey) {
             $element.multiselect('close');
             e.preventDefault();
-          } 
+          }
           else if(e.altKey) {
             switch(e.which) {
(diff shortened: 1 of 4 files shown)


src/jquery.multiselect.filter.js
```diff
@@ -113,47 +113,19 @@
...
-
-        // trigger the change event on the select
-        if(inputCount)
-          $element.trigger('change');
-        
+      // Change the normal _toggleChecked fxn behavior so that when checkAll/uncheckAll
+      // is fired, only the currently displayed filtered inputs are checked
+      var $instance = this.instance;
+      $instance._oldToggleChecked = $instance._toggleChecked;
+      $instance._toggleChecked = function(flag, group, filteredInputs) {
```
mlh758: It looks like the `filteredInputs` param isn't necessary since you're fixing it to true below.

mlh758:
> _Note: The events.js file that I pulled has tabs in it. What I am pushing back has spaces instead. (consider this fair warning...)

I'm tempted to just hit every file with a code formatter before we tag this just to finally get it all consistent. It looks like you changed `bind` to `on` in there, did I miss anything else I should take a second look at?

src/jquery.multiselect.filter.js
SteveTheTechie: Yeah, it could probably be dropped.

SteveTheTechie: Ok just did another commit that removes it.

SteveTheTechie:
> did I miss anything else I should take a second look at?

I think the other thing to note is that the filtering depends on knowing / tracking the hidden state of the inputs in the list.  I am wondering if it may be better to apply/remove a new class name w/ CSS rule display:none rather than directly depending on the state of an intrinsic characteristic?

mlh758:
Heh, I meant in that file. This PR looks good, if you want to change the underlying behavior on how the filtering is done that can be another discussion and PR.

Issue #739, opened 2018-02-02
Truncated menu and how to fix it


SteveTheTechie:
In some cases, the menu can appear truncated.  In particular, you will see this if the multiselect is located in a jQuery UI dialog.   It has to do with the overflow settings of the parent container, what the appendTo is set to, and how it is positioned.

This is an issue that has been addressed a number of years ago by the developers of Chosen (w/ fixes):
https://github.com/harvesthq/chosen/issues/86

SteveTheTechie:
Also relevant:
[https://github.com/harvesthq/chosen/pull/290/commits/a53b58f15a3734ca3affe7c1fe355335aa7d09e8](https://github.com/harvesthq/chosen/pull/290/commits/a53b58f15a3734ca3affe7c1fe355335aa7d09e8)

SteveTheTechie:
The following shows this issue.  It is a small jQuery UI Modal Dialog, and the menu is getting clipped instead of overlaying the dialog like a native select does.
<img width="236" alt="capture-ms-clip" src="https://user-images.githubusercontent.com/5685883/36015325-ea0ea002-0d22-11e8-8c23-6d02090d2b4e.PNG">

SteveTheTechie:
@mlh758 The following shows this issue.  It is a small jQuery UI Modal Dialog, and the menu is getting clipped instead of overlaying the dialog like a native select does.

<img width="236" alt="capture-ms-clip" src="https://user-images.githubusercontent.com/5685883/36015325-ea0ea002-0d22-11e8-8c23-6d02090d2b4e.PNG">

SteveTheTechie:
Closing this issue.

mlh758:
Fascinating, I switched the append behavior to what it is now because it used to pretty much always go to the body. When the widget was set in a dialog it couldn't be interacted with at all, even if it appeared over it.

Could you add this example to the wiki somewhere?

SteveTheTechie:
Sure.  The problem is that we need position:absolute for the menu container so that it is not constrained to the dialog--that is in the CSS already.  The position() _function_ does the magic of positioning the menu to the button, which does reside in the container.   Also, since position:absolute removes the element from the "flow", it does not take up any space in the layout... it is sort of a "free agent", _but only within the constraints of what you are appending to_.

SteveTheTechie:
[...]

PR #749, opened 2018-02-11
Menu clipping

SteveTheTechie:
Addresses #739

1.  Add zIndex option to allow overriding menu z-index for edge use cases.
2.  Add logic to assign a reasonable z-index if not appending to a jQuery UI Dialog.  (e.g., appending to document.body instead)
3.  Go back to setting the menu position on every open in case the button has been moved (e.g. container dialog moved) from the last open.
4.  Close the menu if the mouse cursor is outside the menu and the mouse wheel is turned.   This fixes a scrolling issue that made the menu look detached from the button when the mouse wheel was turned.
5.  Add _allowInteraction support for jQuery modal dialogs to prevent interaction issues when appending to document body to prevent menu truncation.

Note these changes do not directly fix the menu truncation issue.  However, they do make it much easier to address menu truncation by just explicitly setting the appendTo option to document.body, since the z-index would automatically get set to a reasonable value in that case.  The _allowInteraction support is needed for this use case since typically jQuery UI modal dialogs only allow interactions with elements that are contained in the dialog.  (When appending the menu to document.body, typically this would be a case where normal interaction with the menu would be expected to be blocked by the jQuery UI modal dialog code.)

### DIFF
--- a/css/jquery.multiselect.css
+++ b/css/jquery.multiselect.css
@@ -1,7 +1,7 @@
 .ui-multiselect {box-sizing: border-box; padding:2px 0 2px 4px; text-align:left;}
 .ui-multiselect .ui-multiselect-open { float:right }
 
-.ui-multiselect-menu { display:none; box-sizing:border-box; position:absolute; text-align:left; z-index:1010; width:auto; padding:3px;}
+.ui-multiselect-menu { display:none; box-sizing:border-box; position:absolute; text-align:left; z-index:101; width:auto; padding:3px;}
 
 .ui-multiselect-header { display:block; box-sizing:border-box; position:relative; width:auto; padding:3px 0 3px 4px; margin-bottom:3px;}
 .ui-multiselect-header > ul { font-size:0.9em }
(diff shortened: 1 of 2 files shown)


SteveTheTechie:
@mlh758 This is a compromise that lets you "have it both ways".  Essentially, I am not explicitly changing the append behavior as it is fine for the situation where you append to a "tall" dialog where clipping would not occur.  The stacking order has more to do with multiple dialogs overlaying properly, in my view.  I have looked at the jQuery UI code, so I have an idea of what they are doing w/ .ui-front Re: stacking order.

However, while the original code did allow the appendTo element to be explicitly specified for some use cases (e.g. short dialogs), it really did not provide sufficient support for setting the z-index properly for that use case and it did not address modal dialog interaction for that use case.

In essence, this PR does not really change anything for the appendTo ".ui-front use case", as I think that is probably fine for many use cases.  What this PR addresses is the appendTo's that are not '.ui-front'... a reasonable z-index is set for those or you can override it.  Also, the _allowInteraction extension is set for jQuery UI dialogs to prevent interaction problems.

I came to the conclusion in my thinking that there really was no good reliable way to have the code automatically determine whether to appendTo .ui-front or document body., so we should leave that decision to a developer.  We should provide some guidance in a wiki to assist that decision.

SteveTheTechie:
@mlh758 Do you have any specific requests for changes?

mlh758:
No, I've just been really busy the last couple of days and haven't had a chance to finish testing it.

mlh758:
Looks good, merged it.

PR #605, opened 2015-04-17
ISSUE-163 Make optgroups searchable and return its children as well.

luizsgpetri:

PR #751, opened 2018-02-25
New features

### LINKED ISSUE #86: Add keyboard event handlers for closed state in single select mode
In single select mode (multiple: false), when you tab to the closed select menu, you should be able to select options by using the arrow keys without opening the menu. This would replicate the behaviour of the OS select component.

### LINKED ISSUE #163: multiselectfilter plugin ignores optgroup labels
The filter plugin ignores optgroup labels.  If a user types in the group label, I would expect all the items in that group to show up in the filtered results.

### LINKED ISSUE #169: Reopening 'multiple : false' should jump to current selection
When 'multiple : false', the current selection should be fully visible and highlighted upon opening the dropdown.  Currently, the first item is highlighted and not necessarily fully shown - specifically, when there is a scrollbar present, the top of the first item gets chopped off visually.  It would be nice if this widget more closely followed standard Windows UI behavior in regards to the default dropdown.

### LINKED ISSUE #231: Feature Request: Ability to Select/Unselect All/disable programatically for an opt group
```
<select multiselect id="foo">
 <optgroup value="bar">BAZ</optgroup>
 <optgroup value="x">y</optgroup>
 <optgroup value="a">b</optgroup>
 </select>
```

What I want to write is something like this:
<code>
    $('#foo').multiselect('checkAll', {optgroupNum=0}); // checks all options under "BAZ"
</code>

Nice to have:
<code>
$('#foo').multiselect('uncheckAll', {optgroupValue="x"}); // unchecks all options under "y"
$('#foo').multiselect('disable', {optgroupText="b"}); // checks all options under "b"
</code>

Code I currently have to write:
<code>
$($('#foo').multiselect('widget').find('li.ui-multiselect-optgroup-label')[0]).nextUntil('.ui-multiselect-optgroup-label').find(':checkbox').each(function(i, e) {
 if ($(this).attr('aria-selected')) {
  this.click();
 }
});
</code>

### LINKED ISSUE #232: Feature Request: Collapsable opt groups
Would like this as an option on the multiselect.

<code>
  // Make Optgroups collapsible.
  var optgroups = $('li.ui-multiselect-optgroup-label').prepend(
      '<span class="ui-icon ui-icon-triangle-1-s" style="float:left;"></span>');
  optgroups.find('span').click(function() {
    $(this).toggleClass('ui-icon-triangle-1-s ui-icon-triangle-1-e');  // ui-icon-triangle-1-s ui-icon-triangle-1-e from jquery-ui
    var toggle = $(this).hasClass('ui-icon-triangle-1-s');
    $(this).parent().nextUntil('li.ui-multiselect-optgroup-label').toggle(
        toggle);
  });
</code>

### LINKED ISSUE #454: Add the possibility to define own filter rules
At the moment it is only possible to use the build in filter criteria. The used internal regex works as a 'contains' filter. But there can be usecases where a 'beginsWith' filter or other criterias are useful. Because of that there should be the possibility to define own filter rules.

### LINKED ISSUE #467: No automatic scroll when selected option is not visible
Hello,

First of all, thanks a lot for the plugin, it's really amazing!

However, we have found an issue: when the number of options is big and therefore a scroll appears, if the option selected is not visible, there should be an automatic scroll to that option. It is annoying when you want to select the 50th option and right after the 51st one... you need to scroll all the way down again.

See this jsfiddle with the issue:

http://jsfiddle.net/g5r92/1/

Thanks a lot,
Yaiza.

### LINKED ISSUE #513: WAVE toolkit shows up an Accessibility issue with the FIlter box
Using the WAVE toolbar (from FIreFox v24.0)...
WAVE shows a label error on the Filter box of the widget field used to filter the selection.
Can a label be put on the filter box?
Cheers
Jez

### LINKED ISSUE #721: Fixed menu (list box) variant
One of the things I really like about this widget is how it conserves screen space.  It puts a lot of functionality in the space of a button.  

In contrast, the underlying native multi-select takes up more screen space, particularly if there are a lot of options.

However, there are cases involving small set of options where it can be helpful to immediately show the user all of the available options without needing to click a button.   For these situations, it would be helpful to have a variant of this widget that just shows the menu fixed in place (no pop-up) without the button.

### LINKED ISSUE #750: Resync method
A "resync" method would be helpful.  

This would simply resynchronize the selected state between the native select and the widget.

This would be useful since it would allow changing selected states of the native select's options and then this could be reflected on the widget by simply calling the resync method.

Currently, the only way to show this update in the widget is by calling refresh, which rebuilds all the options from scratch which is really overkill if you are just changing selected states as in this case.  Sometimes I think "refresh" should really be called "rebuild" since it is really rebuilding the menu.

Another thing a resync method would make possible would be a val(x) (set value) method for the widget.  This would just call $('select').val(x) on the native select, then call the resync() method to reflect the new selections in the widget. 

```
	 /** 
	  * Updates the widget checkboxes' checked states
	  * from the native select options' selected states.
	  *
	  */
	 resync : function() {
		var $inputs = this.$inputs;
		var $options = this.element.find('option');
		
		if ($inputs.length === $options.length) {
			var inputValues = {};
			$inputs.each( function() {
				inputValues[this.value] = this;				
			});
			$options.each( function() {
				if (this.value in inputValues) {
					inputValues[this.value].checked = this.selected;
				}
			});
			this._trigger('resync');
			this.update();
		}
		else {
			this.refresh();
		}
	 },

```

SteveTheTechie:
**This closes #750, closes #721, closes #605, closes #513, closes #467, closes #454, closes #436, closes #232, closes #231, closes #169, closes #163, and closes #86**

### Many new features and improvements:

1. Expanded on iconSet concept to create linkInfo option to allow configuration of all header link attributes, including icons, text, and titles.  I needed a more general approach to accommodate 2 new link possibilities: Collapse All and Expand All.  (for the opt groups)   The new header option usage provides the means to manage this increased flexibility. (#232)
2. Menu height can now be dynamically resized via the jQuery UI resizable widget, if loaded and resizableMenu option set to true. (#436)
3. Fixed some issues w/ the menu height getting mangled when filtering was in used. (#605)
4. Modifed how header option is used.  Can still set to false to hide header, but now it is also used to specify what header links are seen and in what order. (#232)
5. The multiselect can now also be used as a fixed list box (always open; no pop-up; no button) via the listbox option. (#721)
6. New options groupsSelectable and groupsCollapsable control whether option groups can be selected and/or collapsed, respectively. (#232)
7. Option groups can now be set to be collapsable.  A new small button shows up at the left of each option group heading to collapse/expand the option group. (#232)
8. New resync and value methods allow quickly updating the widget options checked state from the underlying native select without needing to recreate the options via a refresh(). (#750)
9. Can now use the up/down keys on closed single select to change the selected option if the button is focused. (#86)
10. Fixed a previously unknown bug caused by binding menu events to the outer $menu container div instead of the $checkboxes container.  This bug caused the menu events handler to also try to handle header events.
11. Added the ability to reference option groups by their label or index for use as a filter.  Now usable programmatically in addOption, checkAll, uncheckAll, flipAll, collapseAll, and expandAll.  (#231)
12. Simplified the auto width determination for the _setMenuWidth() method by tweaking the related CSS rules.  As a benefit, the _getScrollBarWidth() function has been eliminated.
13. Updated and simplified the _setMenuHeight() method to allow it to better work with filtering. (#605)
14. Updated the open() method so that the first checked option gets the focus when not using the filter. (#169, #467)
15. Re-wrote the filter widget--can now search option group headings and supports alternative filter rules. (#163, #454, #605)
16. Updated and added unit tests.
17. Updated demos and changed theme to Redmond--no more ugly construction orange.
18. Updated i18n files for new linkInfo option format.


mlh758:
By the description this is a lot of welcome changes. I won't be able to look it over today but I'll try to get it merged early this week.

SteveTheTechie:
@mlh758  Understood... I put a lot of time into this, so I would expect that you would need some extra time in to look it over.   I think it is pretty complete, but I will also continue to look for where I may have missed things.  I look forward to any questions you may have as well.

SteveTheTechie:
@mlh758 My intention is that this PR will essentially be my last to contribution for a Version 3 release--I am hopeful that you will feel comfortable doing the release.   😉 

_To that end, I would appreciate it if you would take some time to give this a critical look, not just from whether the feature set, options, methods, etc. make sense, but if there is anything I am missing or could have done better._  

### There were a number of things that I specifically did and did not include in this:

1.  Doing collapsable option groups "right" was a bit more involved than I anticipated.  I needed to enable Collapse All and Expand All links functionality for the header, even though I knew those may not always be needed.  That is what led me to a number of decisions around revamping the header functionality.

2.  The jQuery UI Resizable widget is one of those widgets that I have previously alluded to that does not always work as advertised--it has given me problems in the past in other efforts.  This is why I added an option for enabling its use for dynamic height resizing.

3.  I did more with CSS rules this time around. This worked out really well and allowed me to eliminate some JavaScript.

4.  I implemented your suggestion to create a helper function to find option groups by text or index.  It worked out really well. 👍 

5.  I did not include RTL support...  I do not really know how to do that well, but I am willing to work w/ someone on it in the future, as the gained knowledge could benefit me on my main web app project.

6.  I did not do anything on the open jQuery Validate related issues... I am not familiar w/ jQuery Validate.

7.  I did not do anything with the open ASP related issues... I am not familiar w/ ASP.

8. There are probably a few open issues that I could have checked w/ the new code, but I did not.  I will leave that to you--hopefully, some of those can be closed as being outdated.

9.  I did not do anything related to infinite scrolling, JSON data, or AJAX.  I think that those and the Table View idea I posted as an issue should be done together as a future version update geared to collectively enhancing the widget to support "large data sets" since they all kind of relate to that in one way or another.

10. I did not do anything related to divorcing the code from jQuery UI.  I believe that should be done in a future update.  For some perspective, here is a widget that has jQuery as a dependency, but not jQuery UI (it **used** to depend on jQuery UI also):  https://fullcalendar.io/docs/usage/    _Doing the divorce is doable._

mlh758:
I opened the callbacks demo and tried clicking on a collapsed option group. Instead of selecting everything within it, or doing nothing, it seems to call `uncheckAll` and clear selections. I think it should behave as a normal option group click.

The filter also seems to no longer function in the dialog gist.

src/jquery.multiselect.js
```diff
@@ -19,30 +19,79 @@
...
+         'title': 'Flip all'
+      },
+      'collapse': {
+         'class': 'NOT USED',
+         'icon': '<span class="ui-icon ui-icon-triangle-1-s"></span>',
+         'text': 'NOT USED',
+         'title': 'Collapse'
+      },
+      'expand': {
+         'class': 'NOT USED',
```
mlh758: Do these even need to exist if they're not used?

src/jquery.multiselect.js
```diff
@@ -90,110 +140,130 @@
...
-               :  (typeof headerOn === 'string' ? '<li>' + headerOn + '</li>' : '') );
+      var headerLinksHTML = '';
+      if (options.header === false) {}  // no-op for no header
+      else if (options.header.charAt(0) === '=') {
+         headerLinksHTML = '<li>' + options.header.slice(1) + '</li>';
+      }
+      else {
+         var hdrLinkTemplate = '<li><a class="{{class}}" title="{{title}}">{{icon}}<span>{{text}}</span></a></li>',
+               headerLinks = options.header
+                              .replace(/all\b/g,'All')
```
mlh758: I really don't like that this implementation of headers requires all this string processing/transformation, and a new syntax for `=`. I'd prefer to address the header functionality separately.

This is too much regex and replace on what was an otherwise simple option and the transform to `All` makes me think it has some impact on translations as well.

I'd be a lot happier with something like a mapping object like
```
{
  checkAll: true
  uncheckAll: true
}
```

src/jquery.multiselect.js
```diff
@@ -90,110 +140,130 @@
...
+                              .replace(options.maxSelected ? /\bcheckAll\b/ : '', '')
+                              .replace(/\s|^,+|,+$/g,'')
+                              .split(',');
+         for (var x = 0, len = headerLinks.length; x < len; x++) {
+            if (headerLinks[x] && headerLinks[x] in this.linkInfo
+                && !( headerLinks[x] in {'open':1, 'close':1, 'collapse':1, 'expand':1} ) ) {
+               headerLinksHTML += _linkHTML(hdrLinkTemplate, headerLinks[x]).replace(/<span><\/span/ig, '');
```
mlh758: I'm not okay with the fact that the generated html has to then be fixed with an additional regex replace.

src/jquery.multiselect.js
```diff
@@ -332,21 +400,27 @@
...
           var options = [];
 
           $(elem).children().each( function() {
             options.push(self._makeOption(this));
           });
 
           // Build the list section for this optgroup, complete w/ option inputs...
+          var $collapseButton = !!self.options.groupsCollapsable
+                                 ? $( document.createElement('button') )
+                                    .attr({'style': 'float:left', 'title': self.linkInfo.collapse.title})
```
mlh758: This should probably be a class instead of inline css.

src/jquery.multiselect.js
```diff
@@ -401,6 +488,38 @@
...
+    },
+
+    /**
+     * Updates the widget checkboxes' checked states
+     * from the native select options' selected states.
+     * @param {boolean} skipDisabled If true, disabled options in either are skipped.
+     */
+    resync : function(skipDisabled) {
+      var $inputs = this.$inputs;
+      var $options = this.element.find('options');
```
mlh758: `$options` always appears to be 0, which means this just defers to `refresh` every time.

src/jquery.multiselect.js
```diff
@@ -469,25 +591,47 @@
...
-              case 38: // up
-              case 37: // left
-                self.close();
-                break;
-              case 39: // right
-              case 40: // down
-                self.open();
-                break;
-            }
+          keydown: function(e) {
```
mlh758: Why the switch to keydown?

src/jquery.multiselect.js
```diff
@@ -469,25 +591,47 @@
...
+             // Change selection via up/down on a closed single select.
+				 if (!self._isOpen && !self.element[0].multiple && e.which in {38:1, 40:1} ) {  // up & down keys
```
mlh758: Please don't use in like this, stacking cases are fine.

src/jquery.multiselect.js
```diff
@@ -469,25 +591,47 @@
...
+					var prev, 
+                   current = null, 
+                   next = null;
+					self.$inputs.each( function() {
```
mlh758: This should be a separate method, not a block dropped into the middle of they key handlers.

src/jquery.multiselect.js
```diff
@@ -510,17 +654,24 @@
...
+     * Must be bound to the checkboxes container.
+     * This method scopes actions to filtered options
      * Called by _bindEvents
      */
     _bindMenuEvents: function() {
       var self = this;
+
       // optgroup label toggle support
-      this.$menu.on('click.multiselect', '.ui-multiselect-optgroup a', function(e) {
+      self.$checkboxes.on('click.multiselect', 'a', function(e) {
```
mlh758: If this is now binding to checkboxes instead of menu, the method name should change accordingly.

src/jquery.multiselect.js
```diff
@@ -549,6 +700,34 @@
           checked: nodes.length ? nodes[0].checked : null
         });
       })
+      // collapse button
+      .on('click.multiselect', 'button', function(e) {
```
mlh758: Maybe specify a class for the collapse button so it's more specific to that button instead of any button?

src/jquery.multiselect.js
```diff
@@ -712,21 +894,50 @@
...
+               }
+               self._setMenuHeight(true);
+            }
+         });
+         self.$menu.hide();
+      }
+
+      // Close each widget when clicking on any other element/anywhere else on the page,
+      // another widget instance, or when scrolling w/ the mouse wheel outside the menu button.
+      self.document.on('mousedown.' + self._namespaceID
```
mlh758: These document listeners don't get destroyed when the widget is destroyed, and are created for each widget.

If the widget is destroyed and a new one created, these leak and multiply.

src/jquery.multiselect.js
```diff
@@ -917,12 +1119,12 @@
...
       var overflowSetting = 'hidden';
       var itemCount = 0;
-      var ulHeight = 0;  // Adjustment for hover height included here.
+      var ulHeight = 4;  // Adjustment for hover height included here.
 
       // The following adds up item heights.  If the height sum exceeds the option height or if the number
       //   of item heights summed equal or exceed the native select size attribute, the loop is aborted.
       // If the loop is aborted, this means that the menu must be scrolled to see all the items.
-      $checkboxes.find('li,a').each( function() {
+      self.$checkboxes.find('li:not(.ui-multiselect-optgroup),a').filter(':visible').each( function() {
```
mlh758: Why exclude the option groups here?

mlh758:
This really should have been 10 different PRs, would have been a lot easier on both of us.

Especially with adding #721 in after I asked you not to. I would prefer that as a separate widget like the filter is a separate widget since it changes a lot of behavior.

src/jquery.multiselect.js
SteveTheTechie: Mainly there for documenting so that people do not think they were missed and try to use them.  I can remove them.

src/jquery.multiselect.js
```diff
@@ -90,110 +140,130 @@
...
-      var iconSet = $.extend({}, defaultIcons, options.iconSet || {});
-      var uncheckAllText = options.uncheckAllText;
-      var flipAllText = options.flipAllText;
       var wrapText = options.wrapText || '';
 
+      // Do an extend here to address link info missing from options.linkInfo--missing info defaults to that in linkDefaults.
+      var linkInfo = ( this.linkInfo = $.extend(true, {}, linkDefaults, options.linkInfo || {}) );
+
+      // Helper function used below
+      function _linkHTML(linkTemplate, linkID) {
```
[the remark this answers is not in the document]
SteveTheTechie: It references linkInfo which is defined just above it.

mlh758:
The extra demos gist I made: https://gist.github.com/mlh758/b1b24931db7705441f9786bd68cf3569

One of them is for a dialog, we've used it before for testing/troubleshooting. That's where the speed test thing came from.

SteveTheTechie:
> I opened the callbacks demo and tried clicking on a collapsed option group. Instead of selecting everything within it, or doing nothing, it seems to call uncheckAll and clear selections. I think it should behave as a normal option group click.

Ok, I see that and I agree it is an issue.  

However, there is really more to this.  You have to consider what happens when the filter is in use, also.  The filter was previously set up so that you could not search for optgroup labels, but it would always show all optgroup headings.  If you clicked on a heading, you would only set the checks for the options that were visible / not filtered out.

Now, since I tried to enable searching for option group headings in the spirit of the PRs that requested it, if the optgroup searching option is enabled, it only shows optgroups for which its heading or one of its children contains the search text.  If an optgroup heading matches, then all of its children are shown, also.

So... now there are 3 cases that can show / hide options... for the filter cases, clicking the option group heading is only supposed to change the visible options.   

Consider the case of a filter in use + collapsed option group and someone clicks on the heading...  The way I intended to work was that only visible optgroup options would get change on a heading click, regardless of how they got visible/hidden.  I could change the behavior if a filter is in use vs not... your thoughts?

mlh758:
> but it would always show all optgroup headings.

It only showed option group headings where some children were visible. This meant you could search for something and have results displayed where you just had a few options left under a couple of option groups. Clicking one of those option groups would select only the visible members of that option group.

I think re-ordering the way the filter is done in your code might make this easier, and your addition of the hidden class further simplifies the problem.

1. Apply the search to the options as normal, adding the hidden class to items that should be excluded.
2. If enabled, apply the search to the option groups. If an option group qualifies, remove the hidden class from all of its children.
3. Hide all option groups that have no visible children as before.

It looks like the collapse adds a class to the `ul`, which is separate from the hiding in the filter. It seems like their behavior should be able to overlap safely.

SteveTheTechie:
> It only showed option group headings where some children were visible.

This is incorrect.  Look at line 157 of the filter code for the master branch.

src/jquery.multiselect.filter.js
```diff
@@ -97,98 +124,142 @@
...
+            cache = this.cache,   // Cached text() object
+            optgroupClass = "ui-multiselect-optgroup",
+            hiddenClass = 'ui-multiselect-excluded';
+
+      this.$rows.toggleClass(hiddenClass, !!term);
+      if (!searchGroups) {
+         // If not searching in groups then show all group headings in the results.
+         $checkboxes.find('.' + optgroupClass).removeClass(hiddenClass);
+      }
+      var filteredInputs = $checkboxes.children().map(function(x) {
```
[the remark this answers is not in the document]
SteveTheTechie: These are just index variables...  what would you suggest as an alternative... index1 and index2?

src/jquery.multiselect.js
SteveTheTechie: @mlh758 So do you want me to remove them then?  
I could put the "NOT USED" bit in comments if that would be better.

src/jquery.multiselect.js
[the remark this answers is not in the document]
SteveTheTechie: @mlh758 If you look closely, it uses the linkInfo variable just above it.  I originally had it outside of _create(), but I had to reference this.linkInfo.  Since _create() is the only thing using it, I just decided to pull it inside of _create().  

Would you prefer it be outside of _create()?  Keep in mind it will still need to use the linkInfo stuff somehow.

src/jquery.multiselect.js
SteveTheTechie: @mlh758 I struggled about what to do with this.   It closes a PR, but it is quite a bit of code to just add specialized up/down key functionality for a closed single select.  I could make a separate method, but it will only be used in the keydown handler.

src/jquery.multiselect.js
SteveTheTechie: @mlh758  Ok.  BTW, are you ok w/ a button vs a styled anchor?

src/jquery.multiselect.js
SteveTheTechie: Ok, I will have to address that in the destroy.

src/jquery.multiselect.js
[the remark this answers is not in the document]
mlh758: _create is assigning it to this.linkInfo so the _linkHTML method would be able to access it that way. I would prefer to have it outside the _create method since it isn't a callback, and the data it's capturing in the closure is also accessible via `this`.

src/jquery.multiselect.filter.js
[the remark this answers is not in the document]
mlh758: `groupIndex` and `optionIndex`?

src/jquery.multiselect.js
mlh758: I think you can just remove them, not really a need for a comment either. It seems like their lack of existence would be enough to indicate they aren't used.

src/jquery.multiselect.js
mlh758: The change itself is welcome, It's just a lot of extra code in them middle of an already large, awkwardly indented block. Putting it into a method like `handleButtonKeyboardNav` and just passing that in would be cleaner here.

src/jquery.multiselect.js
mlh758: Yeah I'm okay with it being a button.

SteveTheTechie:
@mlh758 

> This really should have been 10 different PRs, would have been a lot easier on both of us.

My enthusiasm for doing more PRs for this is fading at the moment (it will likely come back later in the future), so I needed to get as much done as I could before I feel the need to move on to other things.

> Especially with adding #721 in after I asked you not to. I would prefer that as a separate widget like the filter is a separate widget since it changes a lot of behavior.

Let's just say that we have a difference of opinion on this.  I am not all that sold on the separate widget approach--it adds unnecessary overhead and perpetuates the use of jQuery UI, which I think needs to be phased out in favor of another class based approach.   Also, #721 only removes the button and fixes the menu position in the html.  It is basically a subtractive approach, and I tend to believe that additional widgets work better for _adding or replacing things, not taking them away_.   If do this as another widget, I would end up having to replace quite a few main widget methods for just some small changes.  Lastly, I believe that the groupColumns option is also a change in behavior/look but that is part of the main widget and not a separate widget--so that does not need to be a separate widget, but this does?

src/jquery.multiselect.js
mlh758: Got it, since find also goes into descendants. Makes sense.

mlh758:
> groupColumns option is also a change in behavior/look but that is part of the main widget and not a separate widget--so that does not need to be a separate widget, but this does?

Touche, leave it.

src/jquery.multiselect.filter.js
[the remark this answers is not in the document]
SteveTheTechie: @mlh758 However, if there are no groups, then groupIndex would actually be the optionIndex.  
(There would only be options.)   Thus, this would be confusing for the no groups case.

src/jquery.multiselect.js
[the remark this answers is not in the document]
SteveTheTechie: Ok, I will move it back outside _create() then.

src/jquery.multiselect.js
SteveTheTechie: @mlh758 ok

src/jquery.multiselect.js
SteveTheTechie: @mlh758 ok will do

src/jquery.multiselect.js
SteveTheTechie: @mlh758 If you have any thoughts on any better ways to account for the heights of the option group headings, I am all ears.  As I noted, I do not think the current method is perfect.

mlh758: I don't at this time, I just wanted to know why the change was made since it was so recently added.

src/jquery.multiselect.js
SteveTheTechie: @mlh758 Ok, so let's unpack this...

> I really don't like that this implementation of headers requires all this string processing/transformation, and a new syntax for =. I'd prefer to address the header functionality separately.

Ok, I wanted to reduce the number of options.   The current header option strikes me as sort of a seldom used option, because the "false" setting is unlikely to be used much.  That is why I thought maybe I could just have values of false and "everything else that implies true" for that option.

>This is too much regex and replace on what was an otherwise simple option and the transform to All makes me think it has some impact on translations as well.

Ok, but you have to keep in mind that I have added three new header link types over what is in the master branch.  I purposely tried to generalize this to avoid continuing down the path of a massive string concatenation operation that would be a big mess.  My thinking is that devs will not likely use all of the header link types but they may want to also specify the order, particularly for a RTL case.

Also, these are just keys for the linkInfo object lookup.   The actual text that would be translated is in the object _values_.

> I'd be a lot happier with something like a mapping object like

I considered that, but you have to keep in mind that unlike arrays, the order of object keys is not guaranteed.   Frequently, they are kept in an internally sorted state to allow for quick look-ups by the JS interpreter.

Because I specifically wanted to allow specifying the order of the header links, I was left with either using an array of keys (for the linkInfo object) or a more convenient comma delimited string of the same that would have to be split--an object is not suitable for this.  The inclusion of a linkInfo key in the array or string implies True, so no object is needed anyway.  The "All" replacement is simply to prevent developer typos from creating issues... since the keys for linkInfo are case sensitive.

I am open to beating this up further, but I am committed to some way of specifying link order to facilitate a future RTL feature and to allow for developer preferences.

src/jquery.multiselect.js
SteveTheTechie: That is just for the case of no text for the link in the linkInfo object.  I could maybe just do a ternary to switch to a different template for the no text case.

src/jquery.multiselect.js
SteveTheTechie: @mlh758  Ok, but I will still have the title attribute there as title cannot be set via CSS.

src/jquery.multiselect.js
SteveTheTechie: Ok... good catch...   It should be `.find('option')`  not  `.find('options')`   Fat fingers got the best of me.  😄

src/jquery.multiselect.js
SteveTheTechie: @mlh758 Please read this:
[https://stackoverflow.com/questions/10191621/jquery-keypress-keydown-which](https://stackoverflow.com/questions/10191621/jquery-keypress-keydown-which)

Long story short, keypress was not working, so I had to switch to keydown.  
If you look at the button keyboard events, I believe it is keydown there also.

BTW, this caught me off-guard.  I had not encountered this difference before.

src/jquery.multiselect.js
mlh758: `_linkHTML` should probably handle that logic instead.

SteveTheTechie:
@mlh758 

> 2. The jQuery UI Resizable widget is one of those widgets that I have previously alluded to that does not always work as advertised--it has given me problems in the past in other efforts. This is why I added an option for enabling its use for dynamic height resizing.

Case in point... try running the Animations demo w/ resizable enabled.

src/jquery.multiselect.js
SteveTheTechie: @mlh758 Ok, I had it there originally, then moved it out.  lol

src/jquery.multiselect.js
mlh758: Adding a sequence field would allow you to still maintain order. Although just saying that it sounds awkward too.

How much of this regex operation is meant to correct for developer typos? It seems like you should just be able to split on commas, map trim over the results, and then continue.

Something like:
```
var headerLinks = options.header.split(',').map(function(headerOption) {
  return headerOption.trim();
}).filter(function(headerOption) {
  if (/open|close|collapse/.test(headerOption)) {
    return false;
  }
  if (options.maxSelected && headerOption === 'checkAll') {
    return false;
  }
  return true;
});
```
It's a little more code, but it's simple and easy to follow, and then the for loop below can just run over the remaining items, removing some complexity from the if check.

SteveTheTechie: @mlh758 Interesting use of filter() w/ map()  ... let me dig into that a bit more later.

The "developer oops" stuff is correcting the "All" case, white space stripping, and extra comma elimination.  I can take that out.

src/jquery.multiselect.filter.js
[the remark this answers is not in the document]
mlh758: Yeah that wouldn't be any good. We could probably reassess what it looks like after the filtering issue is worked out.

src/jquery.multiselect.js
mlh758: Here's another thought. You mentioned the mess of string concatenation...

Would using jQuery to build the template up make it easier/cleaner? Something like is done in makeOption or buildOptionList?

It's another more code/maybe less complicated trade off. As long as `_linkHTML` handles it and produces valid HTML though I'll be happy.

src/jquery.multiselect.js
mlh758: That's okay.

src/jquery.multiselect.js
mlh758: It happens!

src/jquery.multiselect.js
mlh758: Well alright. That is interesting, thanks for changing it then.

src/jquery.multiselect.js
SteveTheTechie: @mlh758 Ok, let me chew on this a bit.   As I am sure you are aware now, I do like speedy code, but sometimes it is worthwhile to strategically sacrifice some speed for cleaner, easier to maintain code.

The string concatenation would likely be fastest.  However, once you get to a certain number of header links, it gets really messy.   I made the decision to try for some cleaner code because building the header up is only done in the _create() "once", not in a loop or anything where you would notice much of a speed difference.  The only place this may matter is if you are doing a lot of widget instances (like I do in my web app), but I think I could live with a small slow-down if it is minimal and makes for cleaner code.

What I would like to see is the following:
1.  Allow the order of header links to be specified and specify which to include.
2.  Have a reasonable way to translate the info in the linkInfo/linkDefaults object into header links without a lot of messy code to do it.  I went with a template approach, but open to alternatives.
3.  Allow for the header to be shut off or custom header text to be specified, but recognize that these are likely edge cases, so maybe they do not have to be as convenient.
4.  Keep the # of options to set up the header to a minimum, if at all possible.

If we can arrive at a way to accomplish the above, I will be fine with that.

SteveTheTechie: @mlh758 One possibility for the header option is to use an array instead of a comma separated string for specifying the displayed links and link order.

Thus, if header === false, no header

If typeof header === array, then process as which links to show and what order,  The array elements are the keys for the linkInfo object to indicate which links and which order.  No string clean up/oops handling needed either.

If typeof header === string, then show custom text.  (No = prefix needed for this case for this approach.)

src/jquery.multiselect.filter.js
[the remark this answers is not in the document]
mlh758: Yeah, leave it as `x` and `y` for now then.

src/jquery.multiselect.js
mlh758: Hey I like that last one a lot. Then we don't need people to learn the '=' syntax, it's just using a string for an absolute value like before. The array for the list let's us skip the split and trim step completely, and anything falsey just shows no header. That's perfect.

SteveTheTechie: @mlh758 Ok, as the array elements would be the keys for the linkInfo object, they would be case-sensitive.  e.g. `['checkall' ,'uncheckall']` will not work.

mlh758: Sounds great. I'd prefer typos cause errors as early as possible anyway.

src/jquery.multiselect.js
SteveTheTechie: Ok, I thought you were using jQuery filter & map functions, but I realized later you are using plain JS array functions.

map & filter are IE9+... _Are you ok w/ that?_

SteveTheTechie: If you want IE8, also, then $.grep & $.map will have to be used.

src/jquery.multiselect.js
SteveTheTechie: @mlh758 I have switched to the falsey/string/array header, and I am using your filter/map approach, also.

However, you still will have to call .html() because of the icons are specified as html elements.  I could build the links up w/ jQuery to get rid of the _linkHTML() function, but I will still be calling html anyway.  Because of this, I sort of with with the template approach.  However, I am opened to being convinced otherwise.  ;-)

src/jquery.multiselect.js
SteveTheTechie: Looking at the following below... doesn't `this.document.off(this._namespaceID);`  get rid of the document listeners for this widget's namespace?
```
    destroy: function() {
      // remove classes + data
      $.Widget.prototype.destroy.call(this);

      // unbind events
      this.document.off(this._namespaceID);
      $(this.element[0].form).off(this._namespaceID);

      if (!this.options.listbox) {
         this.$button.remove();
      }
      this.$menu.remove();
      this.element.show();

      return this;
    },

```

src/jquery.multiselect.js
SteveTheTechie: The main potential problem here is if someone is using collapsable groups, then you also have the height of the collapse buttons to consider.  If someone changes the CSS to make the buttons taller, then potentially the button heights should be examined instead of the anchor heights.   Right now, I am not seeing much of a difference, but it has occurred to me that if the element being examined is an anchor, maybe I should look for a sibling button and use the Math.max height of the two (button and anchor).

src/jquery.multiselect.js
mlh758: I specified IE10+ as a requirement when I released 2.0.

src/jquery.multiselect.js
mlh758: That's what it is supposed to be doing, and in master it appears to be (I just double checked). I wonder if another change along the way broke that?

mlh758:
Just a heads up, tomorrow I'll be leaving on a short vacation so I won't be responding much for the next 5 days.

SteveTheTechie:
Ok thx  I will probably post an update to this PR by Sunday.

src/jquery.multiselect.js
SteveTheTechie: It looks like I was removing the initial period from the namespace ID and it is needed for off().

SteveTheTechie:
@mlh758 I know you are on vacation... Let me know if the changes I made will address the

SteveTheTechie:
@mlh758 Oops closed this by mistake.

mlh758:
> Apparently, widget instance() method is no longer supported, per StackOverflow thread.  Using .data('ech-multiselect') to fetch the instance now--this works w/ the dialog.

Welp, that means this wasn't is my fault. Can you link that Stack Overflow thread? `instance` was added in newer jQuery UI so it not being supported would be very strange. The issue should be that my dialog demo doesn't pull in a new enough jQuery UI.

src/jquery.multiselect.filter.js
```diff
@@ -258,7 +256,7 @@
     */
     destroy: function() {
       $.Widget.prototype.destroy.call(this);
-      this.$input.val('').trigger("keyup");
+      this.$input.val('').trigger("keyup").off('keydown input search');
```
mlh758: Good call here.

src/jquery.multiselect.js
```diff
@@ -598,82 +599,78 @@
...
           },
           mouseleave: function() {
-            $button.removeClass('ui-state-hover');
+            this.classList.remove('ui-state-hover');
           },
           focus: function() {
-            if (!$button.hasClass('ui-state-disabled')) {
-              $button.addClass('ui-state-focus');
+            if (!this.classList.contains('ui-state-disabled')) {
+              this.classList.add('ui-state-focus');
```
mlh758: Cool, I didn't know this was available in IE10.

mlh758:
Code wise I'm happy. My only hangup is wanting to figure out what happened with the `instance` thing, although since that was the only new jQuery UI feature in place, it's probably best to leave it out anyway. No reason to force people to upgrade for one method.

I haven't had a chance to pull this down and test it, but I don't see any more code changes I want.

src/jquery.multiselect.js
SteveTheTechie: Per https://developer.mozilla.org/en-US/docs/Web/API/Element/classList , "Basic Support" is available... Note that I am not using anything fancy w/ it.

SteveTheTechie:
Unfortunately, my computer crashed last night that I have been doing these updates on.   Laptop screen is toast.  It is an old laptop, so I am not sure what I am going to do at the moment.  (typing this on work computer... cannot us it for updates)

All I can say on the instance thing was I kept getting an error of "no such method 'instance' in multiselect instance" in multiselectfilter.js line## ... so I had to figure out a different approach and I did some research on StackOverflow to find the data() approach.

One thing to look at is that I think I put in something like this.element.multiselect().data('ech-multiselect'), but in hindsight, I do not think the multiselect() part should be needed.  (just use this.element.data('ech-multiselect'))

mlh758:
Yup, that was caused by me pulling in an old version of jQuery UI for that demo. The filter thing was my fault. I'll leave your change in though since there's no reason to force an upgrade over a single method that is only convenience. If I find any issues in testing I'll figure out something else to do to fix things up but this will probably be fine to merge at this point.

SteveTheTechie:
Here are the relevant posts, although I guess not needed now as problem solved:
https://stackoverflow.com/questions/30365669/jquery-autocomplete-custom-data-error-no-such-method-instance-for-autocomplete
https://forum.jquery.com/topic/issue-with-jquery-ui-selectmenu-no-such-method-instance

src/jquery.multiselect.js
```diff
@@ -83,27 +77,26 @@
...
+      noneSelectedText: 'Select options', // (string | null) The text to show in the button where nothing is selected.  Set to null to use the native select's placeholder text.
+      selectedText: '# of # selected',    // (string) A "template" that indicates how to show the count of selections in the button.  The "#'s" are replaced by the selection count & option count.
+      selectedList: 0,                    // (integer) The actual list selections will be shown in the button when the count of selections is <= than this number.
+      selectedListSeparator: ', ',        // (string) This allows customization of the list separator.  Use ',<br/>' to make the button grow vertically showing 1 selection per line.
+      maxSelected: null,                  // (integer | null)  If selected count > maxSelected, then message is displayed, and new selection is undone.
+      openEffect: null,                   // (array) An array containing menu opening effect information.
+      closeEffect: null,                  // (array) An array containing menu closing effect information.
```
SteveTheTechie: I changed these option names to be more clear about what they actually do.  However, I also tend to think that the widget may be relying on deprecated jQuery UI API signatures for these, as current docs mention the use of objects for animations.   Therefore, these need to be re-looked at to see if they should be updated to reflect the current API.  I will post an issue for this.

SteveTheTechie:
@mlh758 Are you still testing this?  Are there any further changes needed?

mlh758:
I only got back into town yesterday evening. I haven't had a chance to pull it down yet.

SteveTheTechie:
oops sorry 😊

mlh758:
Everything looks good. Going to merge now.

mlh758:
I published version 3 pre-release with minified files attached so people can start messing with it.

PR #761, opened 2018-03-11
Filter Code Refactor & Fix Scroll on Focus Issues

### LINKED ISSUE #195: Internet Explorer scrolls on mouseover of last element
In Internet Explorer, when the mouse is passed over an element in the drop down list that is just barely visible, the element is scrolled in to view.

As far as I can tell from looking at the code, the reason this happens is because the input element is focused in this segment of code:

```
        .delegate('label', 'mouseenter.multiselect', function(){
            if( !$(this).hasClass('ui-state-disabled') ){
                self.labels.removeClass('ui-state-hover');
                $(this).addClass('ui-state-hover').find('input').focus();
            }
        })
```

commenting out that section causes the keyboard input to not work correctly so I assume it's a necessary evil; however, it's very annoying and bad for usability. 

If you know of any way to avoid this or can fix it, that would be great. Please don't hesitate to contact me if you want any more info on the bug.

Thanks!

### LINKED ISSUE #551: click to scroll issue

SteveTheTechie:
### multiselect.filter.js:

1.  Make filterRules a local variable.  (missed this before)
2.  Remove the unneeded multiselect() fxn call to get the widget instance--just use data() by itself.  (forgot to attempt this before)
3.  Save additional class names in variables that can be obfuscated away.
4.  Use ui-multiselect-grouplabel class name instead of finding the anchor tag directly.
5.  Decrease unneeded jQuery fxn calls w/ use of classList methods.

No feature changes.  All unit tests pass.

### multiselect.js

Fixes #195
Fixes #551

This fixes undesired scrolling issues caused by focusing an element that is only partially visible.
All unit tests pass.

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -85,17 +85,17 @@
       position: {},                       // (object) A jQuery UI position object that constrains how the pop-up menu is positioned.
       zIndex: null,                       // (integer) Overrides the z-index set for the menu container.
       classes: '',                        // (string) Classes that you can provide to be applied to the elements making up the widget.
-      header: ['checkAll','uncheckAll'],      // (false | string | array) False, custom string or array indicating which links to show in the header & in what order.
+      header: ['checkAll','uncheckAll'],  // (false | string | array) False, custom string or array indicating which links to show in the header & in what order.
       linkInfo: null,                     // (object | null) Supply an obect of link information to use alternative icons, icon labels, or icon title text.  See linkDefaults above for object structure.
       noneSelectedText: 'Select options', // (string | null) The text to show in the button where nothing is selected.  Set to null to use the native select's placeholder text.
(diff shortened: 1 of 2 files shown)


SteveTheTechie:
I intended this to be two different PRs but I forgot to create another branch in my repo to separate them.

src/jquery.multiselect.js
```diff
@@ -1363,6 +1374,8 @@
       var speed = this.speed;
       var options = this.options;
       var effect = options.openEffect;
+		var scrollX = window.scrollX;
```
mlh758: The spacing seems off here.

mlh758:
That's alright, the changes are easy enough to review at once this time around. It looks like we're back to the tabs/spaces thing though on these changes.

SteveTheTechie:
>  It looks like we're back to the tabs/spaces thing though on these changes.

Oops, sorry about that.  I forgot to check that.  Should be fixed now.

Issue #154, opened 2011-09-25
Refresh method removing optgroups when using multiselect filter.


BSprite:
I noticed that my optgroups were being destroyed after removing select options when using the refresh method.  I found an example in the comments showing this not to be the case.

HOWEVER, if using the multiselect filter addon then the problem can be duplicated.

Check it out: http://jsfiddle.net/9DzuR/

PR #762, opened 2018-03-13
Prevent unwanted hiding of optgroups after refresh

BrendaH:
Fix same bug as was reported in issue #154.
This original bug was fixed in commit 47854a0 (fix for hiding/showing optgroups when optionlist is hidden) but reintroduced in commit a9fcf40 (Option Groups are now ul elements instead of additional list items. )

Description of problem:
When having optgroups and using filtering in the multiselect, the optgroups get hidden after calling refresh. This problem does not occur when no optgroups are used.

On checking the visibility of the items inside the optgroups the wrong check (:visible selector) was used; the items will not be visible yet. However, they are not hidden on purpose with display:none, so checking on that css is the better way of checking for 'visibility'.

### DIFF
--- a/src/jquery.multiselect.filter.js
+++ b/src/jquery.multiselect.filter.js
@@ -175,7 +175,9 @@
       // show/hide optgroups
       $groups.each(function() {
         var $this = $(this);
-        if(!$this.children("li:visible").length) {
+        // check with a function on display:none css instead of using :visible selector because newly created
+        // (on refresh) items are by default not (yet) visible but not hidden on purpose with the display:none.
+        if(!$this.children('li').filter(function () { return $.css(this, "display") !== 'none' }).length) {
           $this.hide();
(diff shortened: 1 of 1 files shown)


SteveTheTechie:
@BrendaH This PR does not align with the new version 3 filter code, which is different.

I would recommend you check out the version 3 code.

cc:  @mlh758

mlh758:
It would however be good to fix it in version 2 for people that aren't prepared to upgrade to version 3.

mlh758:
I updated the minified files and released 2.0.2 with the fix. I also published it on NPM.

BrendaH:
Thank you. :-)

matthijskooijman:
I had a look over the version 3 code, but from a first review, it seems this bug is not present in the new version (I have not done any testing, though).

PR #768, opened 2018-03-16
Update animation support

### LINKED ISSUE #758: Update animation options syntax for current API
The jQuery UI show/hide animation API apparently has changed to be an object based API... no longer an array-based API.

The old signature may still be usable as a deprecated API signature, but the relevant option syntax should be updated to reflect what the current show/hide animation API signature is.

SteveTheTechie:
Closes #758.

Aligns animation options with current .show() & .hide() signatures.

The current array object support for the openEffect and closeEffect options is maintained.
It does align with the current `.show( effect [, options ] [, duration ] [, complete ] )` signature (or for `hide`), for the case of the `[, options ]` and `[, complete ]` parameters being omitted, since the two array parameters are supplied for the effect and duration parameters.

However, support for a plain object of animation settings for the openEffect and closeEffect options is also now supported.  This aligns with the current `.show( options )` signature (or for `hide`), and allows the easing to be specified in the object.

### DIFF
--- a/css/jquery.multiselect.css
+++ b/css/jquery.multiselect.css
@@ -4,7 +4,7 @@
 .ui-multiselect-menu { display:none; box-sizing:border-box; position:absolute; text-align:left; z-index: 101; width:auto; height:auto; padding:3px; }
 .ui-multiselect-menu.ui-multiselect-listbox {position:relative; z-index: 0;}
 
-.ui-multiselect-header { display:block; box-sizing:border-box; position:relative; width:auto; padding:3px 0 3px 4px; margin-bottom: 2px;}
+.ui-multiselect-header { display:block; box-sizing:border-box; position:relative; width:auto; padding:3px 0 3px 4px; margin-bottom:3px;}
 .ui-multiselect-header > ul { font-size:0.9em }
 .ui-multiselect-header li { float:left; margin:0 10px 0 0;}
 .ui-multiselect-header a { text-decoration:none; }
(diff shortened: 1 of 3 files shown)


SteveTheTechie:
@mlh758 Do not merge this yet... just discovered failing height unit test on the hover clipping commit.

mlh758:
On the animations demo the options need to be updated to `openEffect` and `closeEffect`.

This change also seems to introduce a bug in close where if I click the first menu button the menu attaches to the bottom button (an issue I've seen for a while) but if I click away it no longer closes.

Figuring out what the deal is with the menu going to the wrong button has been on my to-do list for a while but I'd prefer not to introduce a new bug on top of it.

```
Uncaught ReferenceError: $menu is not defined
    at t.(anonymous function).(anonymous function).close
```

SteveTheTechie:
The close issue is just a missing this reference... an oversight on my part.  I did not test thoroughly enough.
😉

SteveTheTechie:
> This change also seems to introduce a bug in close where if I click the first menu button the menu attaches to the bottom button (an issue I've seen for a while) 

This has nothing to do with what I did.   It pre-dates my code.

I did some investigation and I know the cause of it though.  The reason for this has to do with the fact that animations are done asynchronously.  Thus, the only way to ensure that positioning is done in the proper point in the open sequence is to run position in the complete() call-back or via .then().  We can retrieve the deferred from the animated show()/hide() and use that to determine when to do the position, width, and height settings.

You cannot position hidden elements, so the position must happen immediately after the menu is visible.  Getting the sequencing right w/ the animation is tricky. 

The code has never properly handled the async nature of animations, to my knowledge.

SteveTheTechie:
https://www.google.com/search?q=jquery+ui+positioning+animations&oq=jquery+ui+positioning+animations&aqs=chrome..69i57.8334j1j7&sourceid=chrome&ie=UTF-8

Check out the stackoverflow threads on this.

mlh758:
> This has nothing to do with what I did. It pre-dates my code.

Sorry, I didn't meant to imply that you had. That's why I had:
> Figuring out what the deal is with the menu going to the wrong button has been on my to-do list for a while but I'd prefer not to introduce a new bug on top of it.

I just wanted the `close` bug addressed.

PR #773, opened 2018-03-16
Fix 771 Only resize if menu open

mlh758:
The fix was to not resize the menu if the widget isn't open.

I also moved the input binds to their own method, a small refactor.

@SteveTheTechie let me know if this might break any of your new code.

### DIFF
--- a/src/jquery.multiselect.filter.js
+++ b/src/jquery.multiselect.filter.js
@@ -81,7 +81,7 @@
       var opts = this.options;
       var $element = this.element;
 
-      // get the multiselect instance -- instance() method no longer supported -- use data()
+      // get the multiselect instance
       this.instance = $element.data('ech-multiselect');
 
       // store header; add filter class so the close/check all/uncheck all links can be positioned correctly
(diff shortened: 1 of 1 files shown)


SteveTheTechie:
Give me a chance to try to reproduce the issue w/ options that are valid.   The header option that was posted in the original issue is invalid.

src/jquery.multiselect.filter.js
```diff
@@ -94,7 +94,43 @@
           type: "search"
         })
         .css({  width: (/\d/.test(opts.width) ? opts.width + 'px' : null) })
```
SteveTheTechie: Please add a semi at the end here.

src/jquery.multiselect.filter.js
```diff
@@ -213,7 +220,7 @@
       if (term) {
          this._trigger('filter', e, filteredInputs);
       }
-      if (!this.instance.options.listbox) {
+      if (!this.instance.options.listbox && this.instance._isOpen) {
```
SteveTheTechie: I think this is fine.

SteveTheTechie:
@mlh758 I took a look and I think this is fine.

src/jquery.multiselect.filter.js
mlh758: Added it.

PR #776, opened 2018-03-16
Fix #770 - Scroll on Focus w/ _traverse()

### LINKED ISSUE #770: Keyboard Traversing Broken
## I'm submitting a ...
  - [x] bug report
  -  feature request
  -  question

### What is the current behavior?
Cannot key up or down beyond current scroll position.

### What is the expected behavior?
Keys should move the focus and scroll the menu.

### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo
_Common options for demos are [jsfiddle](https://jsfiddle.net/) and [codepen](https://codepen.io)_
Just use the down key to move beyond the bottom of the menu.

#761 broke _traverse().   Should not lock the scroll when using keys to traverse..   

### Please tell us about your setup
  
  * Version: 3
  * Browser:  Chrome
  * Other frameworks in use:

SteveTheTechie:
Fix #770 

Added a parameter to allow _traversing() to allow scrolling (unlock scroll position) on input focus when changing focus via keyboard traversing.

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -1381,24 +1381,25 @@
       // set the scroll of the checkbox container
       this.$checkboxes.scrollTop(0);
 
-      // Show the menu, set its dimensions, and position it.
-      $menu.css('display','block');
+      // show the menu, maybe with a speed/effect combo
+      if (!!effect) {
+         if (typeof effect == 'string') {
+            $.fn.show.call($menu, effect, this.speed);
+         }
+         else if (typeof effect == 'object' && effect.constructor == Array) {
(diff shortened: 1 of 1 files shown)


SteveTheTechie:
@mlh758 So the key distinction here is that when you use the up/down keys to move from one option to another, scrolling _should_ be allowed to happen.   However, if you are just hovering the mouse cursor over options, you should not get a scroll on partially visible options unless you click on them to force the focus to the partially visible input... then the focussed input is scrolled completely into view.

mlh758:
This looks good.

PR #777, opened 2018-03-17
Fix #775 - Animation / positioning sequence

### LINKED ISSUE #775: Fix sequencing of animations and positioning
## I'm submitting a ...
  - [x] bug report
  -  feature request
  -  question

### What is the current behavior?
Reference the animations demo.  The positioning of the animated top menu is wrong.  It is positioned to the second bottom widget.

This is likely caused by the asynchronous execution of animations, causing the positioning to actually take place before the animation completes.

You can also see this in the fact that the width of the menu does not match the button width as it should.  Again, the asynchronous animations are messing up the sequencing.

### What is the expected behavior?
Each animated menu should be positioned to their respective widget on open.

### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo
_Common options for demos are [jsfiddle](https://jsfiddle.net/) and [codepen](https://codepen.io)_

Run the animations demo.

### Please tell us about your setup
  
  * Version: 3
  * Browser: Chrome
  * Other frameworks in use:

SteveTheTechie:
Fix #775 

Position / set dimensions of menu before doing animation.
Add this.element.trigger('blur') to close()

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -1389,14 +1389,16 @@
 
       // Do any specified open animation effect after positioning the menu.
       if (!!effect) {
+         // Menu must be hidden for some effects (e.g. fade) to work.
+         $menu.css('display','none');
          if (typeof effect == 'string') {
-            $menu.effect(effect, this.speed);
+            $menu.show(effect, this.speed);
          }
          else if (typeof effect == 'object' && effect.constructor == Array) {
-            $menu.effect(effect[0], effect[1] || this.speed);
+            $menu.show(effect[0], effect[1] || this.speed);
          }
          else if (typeof effect == 'object' && effect.constructor == Object) {
-            $menu.effect(effect);
+            $menu.show(effect);
          }
       }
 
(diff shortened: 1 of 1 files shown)


SteveTheTechie:
@mlh758 So it turns out that you can split off the animation effect from the actual show()ing of elements.  I basically did that so that I could display the menu, then set dimensions, position it, and finally do the animation effect.  If you run the animation demo w/ this, it looks like this does the trick.

SteveTheTechie:
The animation does execute asynchronously, and my original working code had a $.Deferred() in it to manage that.  However, once I figured out that the animation effect stuff could be split from the show()ing part of the code, it became apparent that once I moved the position() call before the effect, what was left after the effect() in the code was not really impacted by the async nature of the animation.  Hence, there is no $.Deferred() in this commit as a result.  If we determine that the async nature of the animation has other impacts, we can put the $.Deferred() back to handle that.  I am experienced w/ $.Deferred(), so I know how to set that up.

mlh758:
Closing this for #790 which is the same PR just rebased and squashed.

PR #783, opened 2018-03-18
Fix #767 - Fix Hover Clipping

SteveTheTechie:
This is a bug fix.

Fixes #767 

Checkboxes UI padding added in css file.
Menu height calcs adjusted to account for padding added.

Unit tests pass.
No new unit tests.
No demos updated.
No i18n files updated.

PR #784, opened 2018-03-18
Use $.outerWidth() for Native Select Element Width

SteveTheTechie:
This is a bug fix.

When the widget is created in the create() event of a dialog, the native select will be hidden and _getBCRWidth() will return a zero width.   Therefore, $.outerWidth() must be used to retrieve the native select's width in _create(), since $.outerWidth() will unhide the select to get the width.


SteveTheTechie:
This should be merged after #783 for best results, as the code is based on that.

SteveTheTechie:
Closed and recreated with just one commit

PR #807, opened 2018-04-08
Update jquery.multiselect.filter.css

SteveTheTechie:
Filter text should not be so tiny.

### This Pull Request is a
 - [x] Bug fix
 - [ ] Feature addition
 - [ ] Code refactoring / optimization
 - [ ] Test Addition
 - [ ] Demo Addition
 - [ ] i18n Addition
 - [ ] Other (explain below)
   
### What changes are you proposing?  Why are they needed?<br>
_(Provide use cases, code references, [jsfiddle](https://jsfiddle.net/), [codepen](https://codepen.io), etc.)_   

Filter text should not be so tiny.
Change to .9em's

  
  - [x] No Tabs / Code Spacing Consistent
  - [x] Tests Ran and 0 Failing
  - [x] Tests Added/Updated
  - [x] Impacted Demos Updated
  - [x] Impacted i18n Code Updated
  
### @Mentions:
@mlh758

### DIFF
--- a/css/jquery.multiselect.filter.css
+++ b/css/jquery.multiselect.filter.css
@@ -1,4 +1,4 @@
 .ui-multiselect-hasfilter ul { position:relative; top:2px }
-.ui-multiselect-filter { float:left; margin-right:10px; font-size:11px; width:100%; }
-.ui-multiselect-filter input { width:100px; font-size:10px; margin-left:5px; height:15px; padding:2px; border:1px solid #292929; -webkit-appearance:textfield; -webkit-box-sizing:content-box; }
+.ui-multiselect-filter { float:left; margin-right:10px; font-size:.9em; width:100%; }
+.ui-multiselect-filter input { width:100px; font-size:.9em; margin-left:5px; height:15px; padding:2px; border:1px solid #292929; -webkit-appearance:textfield; -webkit-box-sizing:content-box; }
 .ui-multiselect-excluded {display: none}


mlh758:
I don't think this or the current setting is actually doing anything. They both appear to be getting overridden to 1em by jquery ui.
image

SteveTheTechie:
hmm let me take another look

SteveTheTechie:
@mlh758  I have figured out what is going on here... So let's unpack this.

1.  First, lets observe that the `.ui widget input` and the `.ui-multiselect-filter input` selectors both apply and are actually equal CSS specificity.  Thus, the last rule parsed by the browser is the one that "takes".

2.  In my HTML, I include the jquery-ui css 1st and the multiselect css after that.  I do this because I consider the jquery-ui css to be more generic and I want the multiselect css that I consider to be more specialized to override any rules where appropriate.  This is specifically geared to a case of equal specificity so that the multiselect css will be the winner in the rule comparison.

3.  However, in the filter demo html that I assume you are referencing, the CSS inclusion order is _reversed_.   This means the last rule parsed for this case of equal specificity would be the one from the jquery-ui css not the one from the multiselect css.

In a nutshell. we can certainly change the order, but in reality, what we probably should do is to up the specificity of the `.ui-multiselect-filter input` rule so that it always wins the comparison regardless of order of CSS rules parsed.

2 fixes here then:

1.  Change the order to include jquery-ui css first.  (more for optics though) 
2.  Change the `.ui-multiselect-filter input` selector to be `.ui-widget .ui-multiselect-filter input` in the rule instead.

SteveTheTechie:
I will do another different PR after this one to reorder the CSS loading in the demos so that the jquery-ui css loads first.  It is loading last after the other CSS in all the demos that I see.

PR #818, opened 2018-05-15
Auto scroll to first selected item in single mode

psyafter:
When menu opened - scroll to first selected item

### This Pull Request is a
 - [ ] Bug fix
 - [V] Feature addition
 - [ ] Code refactoring / optimization
 - [ ] Test Addition
 - [ ] Demo Addition
 - [ ] i18n Addition
 - [ ] Other (explain below)
   
### What changes are you proposing?  Why are they needed?<br>
_(Provide use cases, code references, [jsfiddle](https://jsfiddle.net/), [codepen](https://codepen.io), etc.)_   

  
  - [ ] No Tabs / Code Spacing Consistent
  - [ ] Tests Ran and 0 Failing
  - [ ] Tests Added/Updated
  - [ ] Impacted Demos Updated

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -763,12 +763,18 @@
       this.position();
 
 
-      // select the first not disabled option or the filter input if available
+      // Set focus to the first selected (in single mode) or not disabled option or the filter input if available
+      var $firstSelected = !o.multiple ? $container.find('li>.ui-state-active').first() : [];
+      if ($firstSelected.length) {
+        $firstSelected.trigger('mouseover').trigger('mouseenter').find('input').trigger('focus');
+      }
       var filter = this.header.find(".ui-multiselect-filter");
       if(filter.length) {
         filter.first().find('input').trigger('focus');
       } else if(this.labels.length){
-        this.labels.filter(':not(.ui-state-disabled)').eq(0).trigger('mouseover').trigger('mouseenter').find('input').trigger('focus');
+        if (!$firstSelected.length) {
+          this.labels.filter(':not(.ui-state-disabled)').eq(0).trigger('mouseover').trigger('mouseenter').find('input').trigger('focus');
+        }
       } else {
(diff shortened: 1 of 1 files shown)


psyafter:
added fix for "with filter" mode

mlh758:
This still doesn't seem to work as described. If I open the single select demo and add more elements to it (ensuring I have to scroll) and pick the last item, close the menu, reopen the menu, the first element is focused and in view. It does not scroll to the last element.

In filtered mode, it seems to scroll to the last element the first time, as long as I don't mouse over anything else first. Subsequent openings go to whatever element got focused by the mouse as I moved it across the menu rather than what is selected. I don't like that inconsistency.

psyafter:
Did you checked right patch?
https://github.com/psyafter/jquery-ui-multiselect-widget/tree/patch-2
I retested and both working well.

The need for it is to continue selecting from same point (last selection). For example you have callback "onchange" and do something on selection changes. After, you opening selectbox and want to continue selecting, but it make it hard if each time a menu focus goes to first item in list instead to focus on selected item, so you can continue with mouse or keyboard from last selected item and not lose time for scrolling

mlh758:
I know what went wrong, I was testing with the demo pages which I just converted to use a minified version of the code for GitHub Pages. Your changes weren't in place. I'll test again today.

mlh758:
Okay, yeah, I see the changes as working. Would you be able to squash your commits and update with master?

After that I'll go ahead and merge your changes in.

psyafter:
why with master?


______________________
Sasha Bernstein

mlh758:
Sorry, at work we mostly just rebase our branches so the merge is a simple fast forward. Makes for a simple history. That's probably not how people usually do it on GitHub.

I'd still like you to squash your changes.

psyafter:
done

mlh758:
Merged. Thanks for contributing.

PR #824, opened 2018-06-28
Add an boolen option multipleFilter , use a,b to search a or b with true value

Gavinluo:
Add an boolen option multipleFilter , use a,b to search a or b with true value

in my project , i want to use [ , ] to split search words to filter content , so i add this branch, and want to share to every one , please approved , thanks.

this is my first pull request , ^_^


mlh758:
Hi, I appreciate the PR but I don't think this is a change I would like to bring into the main repository.

PR #825, opened 2018-08-03
fix _toggleDisabled error this.$button is undefined when listbox option is set to true

Mazenrab:
This Pull Request is a bug fix

Check existing of $button inside _toggleDisabled function becase $button is undefined if listbox option is set to true.

  - [x] Tests Ran and 0 Failing
  - [ ] Tests Added/Updated
  - [ ] Impacted Demos Updated


mlh758:
Thank you for contributing.

PR #828, opened 2018-08-27
Fixes height calculation for optgroups

fastcatch:
Height calculation for optgroups have been slightly off as optgroup CSS properties are different from those of items and it has not been considered properly.

  - [ X ] Tests Ran and 0 Failing
  - [ - ] Tests Added/Updated
  - [ - ] Impacted Demos Updated
  - [ - ] Impacted i18n Code Updated

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -585,7 +585,10 @@
     _setMenuHeight: function() {
       var headerHeight = this.menu.children(".ui-multiselect-header:visible").outerHeight(true);
       var ulHeight = 0;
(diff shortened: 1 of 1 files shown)


mlh758:
What is the visual difference I should be looking for with this change? I merged this code on my local and fired up the demo pages and it looks the same as master. Where were you seeing a problem?

fastcatch:
Sorry for the misnamed variable, I've fixed it.

The problem does not, indeed, show up in the test.  However, it does exist, see the following two images:

optgroup_bad

optgroup_ok

I suspect it has to do with theming but I haven't investigated it deeply.

mlh758:
Alright, that makes sense. Just wanted to make sure the change was fixing something useful. I'll merge this now.

Issue #837, opened 2019-03-11
For some reason, the dropdown menu gets placed in a wrong position


rasben:
Im 99% sure that this is not an actual issue with the library, but i hope someone here recognises my issues and can nudge me in the right direction

### What is the current behavior?
- I init the multiselect widget and it looks great
- I click on the element
- The dropdown menu gets placed totally randomly on my page

<img width="749" alt="Screenshot 2019-03-11 at 16 18 19" src="https://user-images.githubusercontent.com/12376583/54150862-cfb3c300-4439-11e9-8adc-eb6cb027b6fd.png">

### What is the expected behavior?
- I would expect the dropdown to appear below the input

### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo
- I've tried so hard to reproduce it in a codepen, but it just doesnt work
- I *think* its to do with margins/paddings on the page.
- When i disable some of my CSS, the element works fine

### Please tell us about your setup
  
  * Version: v2 and v3 
  * Browser: Chrome
  * Other frameworks in use: Drupal

mlh758:
If you call the `refresh` method on it, does it end up in the right place?

rasben:
@mlh758 you mean like this?

```
$("#the_id_of_my_original_select").multiselect("refresh")
```

i tried to run that in my chrome console:

<img width="291" alt="Screenshot 2019-03-11 at 23 11 45" src="https://user-images.githubusercontent.com/12376583/54161637-0ea24280-4453-11e9-9786-47172233d77b.png">


and i still have the same issue :(

mlh758:
I wrongly assumed that `refresh` would also reset the position of the widget.

`$("#the_id_of_my_original_select").multiselect("position")` should actually impact the position.

I'm wondering if something else on the page is interfering with where the widget is placing the menu. If resetting the position manually like this helps we at least know where to start looking.

rasben:
same result with .multiselect("position")

I know for sure that it's some CSS on the page that's causing troubles, but I really cannot figure out what part of it that causes the trouble

I was hoping someone had run into the same problem and could say "Oh, its cause you used float on the parent" or something :(

rasben:
Hmm.. Ok after digging a bit, i found the following "solution":
- the Drupal admin adds a sticky toolbar header. To leave space for that, It also adds a padding-top value on the `<body>`.
[...]

PR #838, opened 2019-03-13
ISSUE 837: Use jquery offset instead of .position().

rasben:
As i describe in my issue, my menu gets placed in a strange random position.
I found out that the issue is with the jQuery `.position()` function, which for some reason gets confused on my setup.
Both the `left` and `top` properties are wrong - if I base it on the `.offset()` functionality, then it works fine

I have a feeling that this fix can have some pretty crazy consequences - and it really surprises me that noone else seems to have this issue..

Screenshot 2019-03-11 at 16 18 19

### Related Issue numbers 
https://github.com/ehynds/jquery-ui-multiselect-widget/issues/837

  - [x] Tests Ran and 0 Failing
>"Took 821 ms to run 334 tests. 334 passed, 0 failed."
  - [ ] Tests Added/Updated
  - [ ] Impacted Demos Updated

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -1506,8 +1506,11 @@
         this.$menu.position(this.options.position);
       }
       else {
-        var pos = $button.position();
-        pos.top += this._savedButtonHeight;
+        var pos = {};
+
+        pos.top = $button.offset().top + this._savedButtonHeight;
+        pos.left = $button.offset().left;
+
         this.$menu.offset(pos);
       }
     },


rasben:
> Pull Request Approval Checklist:
> Tests Ran and 0 Failing
> Tests Added/Updated
> Impacted Demos Updated
> Impacted i18n Code Updated

Is this something I need to do, or is it the maintainers job? If I need to do it, how do I run these tests? :)

mlh758:
You should base your changes off of the version 3 branch, I'm trying to keep all future changes there (and I should really just merge it at some point).

Using the version3 branch doing the tests is pretty easy. The `package.json` has has a script in place so just `npm install` and `npm run test`.

I'm going to look through the commit history though, I have some memory of explicitly moving away from offset to use position due to some other error.

mlh758:
Never mind on the position thing, that was somewhere else. I think this should be good to merge once you've verified the tests and made the changes to version3 instead of master. The code should be pretty similar.

Thank you for contributing.

mlh758:
Your changes fix the menu going to the wrong widget in the position demo 👍

rasben:
Ok great - I'll do those things and get back to you, probably today, maybe tomorrow

Thanks!

rasben:
Ok Ive updated/commited/pushed and changed against V3 @mlh758 :)

I ran the tests and they didnt cause any problems:
>Took 821 ms to run 334 tests. 334 passed, 0 failed.

Ruby is a bit fucked up on my machine right now, so I cant test the demo site sadly - could you perhaps check? :)

As mentioned before (and i guess as you know), this is a change that works fine for me, but I am a little worried that it would break lots of scenarios out there for people, for god knows what reasons - Im not even sure why the original is an issue on my setup after all..

I guess seeing as version3 is still in pre-release, its at their own peril to update though :)

PR #841, opened 2019-03-14
Introduce state class for when any item is selected.

rasben:
Adding a class to the button when any value is selected.

The usecase of this for me is so I can style the button to have a "placeholder" styling.
Technically it's not a placeholder, but you get the point :)

  - [x] Tests Ran and 0 Failing
>Took 1320 ms to run 334 tests. 334 passed, 0 failed.

  - [ ] Tests Added/Updated
  - [ ] Impacted Demos Updated

### DIFF
--- a/src/jquery.multiselect.js
+++ b/src/jquery.multiselect.js
@@ -648,6 +648,8 @@
       var value;
 
       if (numChecked) {
+        this.$button[0].classList.add('ui-has-selected-value');
+
         if (typeof selectedText === 'function') {
           value = selectedText.call(this, numChecked, inputCount, $checked.get());
         }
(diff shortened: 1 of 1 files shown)


rasben:
Here's an example of how I'm using it:
Screenshot 2019-03-14 at 19 33 14

The one on the left is selected state, the one on the right is unselected

mlh758:
This looks fine to me. Could you add a test for it?

rasben:
Sure, but can you nudge me in the right direction of where/how I'd add that test? @mlh758 :)

mlh758:
I think an easy way would be to add another expectation to the `checkAll` method [here](https://github.com/ehynds/jquery-ui-multiselect-widget/blob/version3/tests/unit/methods.js#L114).

Just make sure your class is set on it. Then on `uncheckAll` make sure your class is no longer set on it.

PR #864, opened 2020-04-03
bugfix

fglueck:
- fixing wrong $elemnent access

if noneSelectedText set to '' it throw a js error: $element not exists...


mlh758:
Looks good to me.

